PR #44, opened 2024-11-18
[PLD] Sheltron Overcap Health Check optional

edewen:
A forced health check was added to overcap shelltron a few months back. Simply made it an option for 2 reasons. 

Shelltron is a mit more than a heal. I would rather it prevent the hit than be used right after it 

Forcing a health check prevents something like the below reaction for using overcap as an offtank. 

image


edewen:
Closing as kaeris talkign with me about adding intervention feature instead.

PR #56, opened 2024-11-19
[MCH] fix hypercharge weave window

Kagekazu:
- fix hypercharge weave window
- fix obselete value


Kagekazu:
Think i pushed an ecommons update by mistake - sorry Tauren 😄

PR #102, opened 2024-11-23
[Core] `/wrath` auto-rotation enable and disable commands fix

erdelf:
fix for https://github.com/PunishXIV/WrathCombo/pull/63

the command switch only operates on the first word, not on multiple.

### DIFF
--- a/XIVSlothCombo/XIVSlothCombo.cs
+++ b/XIVSlothCombo/XIVSlothCombo.cs
@@ -623,28 +623,10 @@ namespace XIVSlothCombo
                     }
                 case "auto":
                     {
-                        Service.Configuration.RotationConfig.Enabled = !Service.Configuration.RotationConfig.Enabled;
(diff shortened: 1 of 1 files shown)


XIVSlothCombo/XIVSlothCombo.cs
```diff
@@ -623,29 +623,11 @@ private void OnCommand(string command, string arguments)
                     }
                 case "auto":
                     {
-                        Service.Configuration.RotationConfig.Enabled = !Service.Configuration.RotationConfig.Enabled;
+                        Service.Configuration.RotationConfig.Enabled = argumentsParts.Length > 1 ? argumentsParts[1].ToLower() == "on" : !Service.Configuration.RotationConfig.Enabled;
```
zbee, line 626: Great way to do this instead, very clean.

zbee (approved):
Great catch, my testing was clearly flawed.

PR #107, opened 2024-11-24
[PVP] Various small pvp fixes Work in progress

edewen:
DRK
Fixed the wrong buff code on blackest night, it was overwriting itself. Added was last ability used as good measure bc of server ticks.
Move the shadowbringer health slider to where it should be in config screen.

### DIFF
--- a/XIVSlothCombo/Combos/PvE/DRK/DRK_Config.cs
+++ b/XIVSlothCombo/Combos/PvE/DRK/DRK_Config.cs
@@ -176,7 +176,7 @@ internal partial class DRK
 
                 #region PVP
 
-                case CustomComboPreset.DRKPvP_Burst:
+                case CustomComboPreset.DRKPvP_Shadowbringer:
                     UserConfig.DrawSliderInt(1, 100,
                         DRKPvP.Config.ShadowbringerThreshold,
                         "HP% to be at or Above to use " +
--- a/XIVSlothCombo/Combos/PvP/DRKPVP.cs
+++ b/XIVSlothCombo/Combos/PvP/DRKPVP.cs
@@ -21,7 +21,7 @@ namespace XIVSlothCombo.Combos.PvP
         {
             public const ushort
                 Blackblood = 3033,
-                BlackestNight = 1038,
+                BlackestNight = 1308,
                 SaltedEarthDMG = 3036,
                 SaltedEarthDEF = 3037,
                 DarkArts = 3034,
@@ -60,7 +60,7 @@ namespace XIVSlothCombo.Combos.PvP
 
                         if (canWeave)
                         {
-                            if (IsEnabled(CustomComboPreset.DRKPvP_BlackestNight) && ActionReady(BlackestNight) && !HasEffect(Buffs.BlackestNight))
+                            if (IsEnabled(CustomComboPreset.DRKPvP_BlackestNight) && ActionReady(BlackestNight) && !HasEffect(Buffs.BlackestNight) && !WasLastAbility(BlackestNight))
                                 return OriginalHook(BlackestNight);
 
                             if (IsEnabled(CustomComboPreset.DRKPvP_SaltedEarth) && ActionReady(SaltedEarth) && IsEnabled(CustomComboPreset.DRKPvP_SaltedEarth))


WrathCombo/Combos/PvE/DRK/DRK_Config.cs
```diff
@@ -176,7 +176,7 @@ internal static void Draw(CustomComboPreset preset)
 
                 #region PVP
 
-                case CustomComboPreset.DRKPvP_Burst:
+                case CustomComboPreset.DRKPvP_Shadowbringer:
```
Kaeris-Tempest, line 179: Shadowbringer requires a minimum of 12,000 HP to use (or Dark Arts to be present).
The logic will softlock if your HP drops below that, so you need a minimum HP check.

WrathCombo/Combos/PvP/DRKPVP.cs
```diff
@@ -60,7 +60,7 @@ protected override uint Invoke(uint actionID, uint lastComboActionID, float comb
 
                         if (canWeave)
                         {
-                            if (IsEnabled(CustomComboPreset.DRKPvP_BlackestNight) && ActionReady(BlackestNight) && !HasEffect(Buffs.BlackestNight))
+                            if (IsEnabled(CustomComboPreset.DRKPvP_BlackestNight) && ActionReady(BlackestNight) && !HasEffect(Buffs.BlackestNight) && !WasLastAbility(BlackestNight))
```
Kaeris-Tempest, line 63: I'd suggest replacing `WasLastAbility` with `WasLastAction` or it won't use consecutive Blackest Nights between weaponskills.

WrathCombo/Combos/PvP/GNBPVP.cs
```diff
@@ -47,19 +48,29 @@ internal const ushort
...
+            
             protected override uint Invoke(uint actionID, uint lastComboMove, float comboTime, byte level)
             {
                 if (actionID is KeenEdge or BrutalShell or SolidBarrel)
                 {
+                    int corundumThreshold = GetOptionValue(Config.corundumThreshold);
+                    bool enemyGuard = TargetHasEffect(PvPCommon.Buffs.Guard); //Guard check
+                    bool inGF = JustUsed(GnashingFang, 3f) || JustUsed(SavageClaw, 3f) || JustUsed(WickedTalon, 2f);
+
+                    if (CanWeave(ActionWatching.LastWeaponskill) && IsEnabled(CustomComboPreset.GNBPvP_Corundum) && PlayerHealthPercentageHp() <= corundumThreshold && IsOffCooldown(HeartOfCorundum))
```
Kaeris-Tempest, line 71: Looks functional, but I would highly advise against using `canWeave` here (you're just throwing away free damage from Continuation procs). Let them go through even if it's not in a weave window.
PvP is far too chaotic, and unlike PvE, you cannot guarantee weaves.

WrathCombo/Combos/PvP/DRGPVP.cs
```diff
@@ -79,6 +80,9 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                         if (IsEnabled(CustomComboPreset.DRGPvP_WyrmwindThrust) && HasEffect(Buffs.FirstmindsFocus) && GetTargetDistance() >= GetOptionValue(Config.DRGPvP_Distance_Threshold))
                             return WyrmwindThrust;
                     }
+
+                    if (IsEnabled(CustomComboPreset.DRGPvP_ChaoticSpringExecute) && IsOffCooldown(ChaoticSpring) && EnemyHealthCurrentHp() <= 8000)
+                            return ChaoticSpring;
```
Kaeris-Tempest, line 85: Just some food for thought as a DRG player.
Chaotic Spring also has a healing component (and a strong one at that).
I personally would give it two conditions: target's health is low **OR** your health is low.
If you're feeling frisky, you can also add a `hasLifeOfTheDragon` condition to ensure boosted damage.

WrathCombo/Combos/PvP/DRKPVP.cs
edewen: TBN is an ability. You dont want to use it twice in a single gcd. it is overwriting itself. it literally fixing that because some server ticks are bad

WrathCombo/Combos/PvE/DRK/DRK_Config.cs
edewen: Makes sense will implement

WrathCombo/Combos/PvP/DRGPVP.cs
edewen: image
I added and execute, but the sustain for health section is still there higher up. and the execute is tied to the amount of damage it would do. 8k

WrathCombo/Combos/PvP/GNBPVP.cs
Kaeris-Tempest: Happened to look at a different block at the same time. This is fine.

PR #111, opened 2024-11-24
[RDM] Melee combo Fill Option and other features.

nativepapaya:
**Update**: All features finalized. Currently testing. Any feedback is appreciated.

**Update 2:** I believe I've fixed all outstanding issues.

**_Features_**

1.  Add overcap protection as its own option **DONE**
2. Add melee fill option **DONE**
3. Add the option to hold one Acceleration charge for manual use **DONE**

**Preview of features:**
Melee fill option and overcap protection
image

Acceleration hold charge
image

----------------------------------------------------------------

Adds a fill option for the RDM melee combo similar to DNC dance steps. The melee combo will currently execute on the main rotation when either of the 3 following conditions are true:

- User manually initializes the combo using Enchanted Riposte (if fill is selected)
- Embolden and Manafication are off cooldown (If option is selected)
- Black and White mana reach the overcap protection value.

~~The overcap protection is currently hard coded but can be made into an option where users can select how much mana to hold before the combo is initialized automatically. This can also be made into a separate feature entirely not dependent on the "fill" option.~~
**FIXED**

**Current issues/Concerns**
~~There is currently a bug where if you tick the "Use Manafication and Embolden Option" then untick, it will still registered as selected until you reload the plugin. I'm not sure if this is a bug because of the way I wrote the feature or something else.~~
*FIXED*

~~The current implementation for checking whether or not the user has manually executed the combo is kind of hacky and I am hoping to find a better solution. The way it currently work is, I have a global bool that is set to true when the user manually executes Enchanted Riposte. It then will be set to false once a combo finisher is used. It works as intended (so far) but I am sure it could be made better. I was searching to see if there is anything related to RDM that tracks whether or not you are currently performing a melee combo.~~
*FIXED*

I am also not currently level 100 so not sure how this feature will work with the balance opener. Since openers are hard coded it shouldn't be an issue but if someone that has a 100 RDM can confirm, it be great.


nativepapaya:
I also noticed that there is currently a check for how much mana you have before using Manafication. This is probably a leftover from last expansion as Manafication no longer gives you mana and can be used whenever you don't have 3 mana stacks.

Tartarga:
> I also noticed that there is currently a check for how much mana you have before using Manafication when the double option is not selected. This is probably a leftover from last expansion as Manafication no longer gives you mana and can be used whenever you don't have 3 mana stacks.

RDMMana handles this Dawntrail's change to Manafication.

nativepapaya:
I improved the way the check for being in a melee combo is handled. Basically just check if the last weapon skill was either Enchanted Riposte or Enchanted Zwercchau. The previous solution, outside looking plain bad, also didn't work when synced below a level were you had a finisher. This fixes that issue.

Tartarga:
For the sake of transparency, I mostly inherited this code as I play RDM after SGE. I understand most of it with the exception of the Manafication burst code. I do not to intend to sound harsh or come off elitest so if I do I apologize in advanced. C# oddly as it

I'm trying to understand what the current problem in Wrath you are aiming to address. I see 2 items of importance. The first being over cap protection (valid) and the Manafication and Embolden issue (not sure if that's your current code or wrath). Please ELI5

nativepapaya:
Might be best though to make "overcap", "burst mode", and "fill" all separate options instead of just rolling them all into one like I currently have it. That way the user has more control over the melee combo.

nativepapaya:
Added overcap protection that is independent of the melee "fill" option.

Preview:
image

nativepapaya:
Also had the idea of manually keeping a charge of Acceleration for manual use. This can be saved for movement. What do you think about this @Tartarga?

nativepapaya:
Been testing these changes and so far everything seems to be working.

Tartarga:
So zbee helped me decode what the goal was, and I mostly get the goal now, but I'm curious about a different approach.

These 3 are your requirements to start the melee combo, correct?

> - User manually initializes the combo using Enchanted Riposte (if fill is selected)
> - Embolden and Manafication are off cooldown (If option is selected)  
> - Black and White mana reach the overcap protection value.

First Req: I believe I have your first requirement easy peasy in PR #179 I'd like you to check. 
Second Req: Part of me also thinks, that last bit I changed, can be expanded for quick check for Embolden & Manafication buffs, but then the description needs to be changed to say that Riposte or both buffs.
Third Req: As a separate option, I would think this would be best bled off with Reprise, which would be an option outside of Melee Combo?

With Also with Second Req I think it would be incompatible with Hold for Double, as it does the holding.

nativepapaya:
> So zbee helped me decode what the goal was, and I mostly get the goal now, but I'm curious about a different approach.
> 
> These 3 are your requirements to start the melee combo, correct?
> 
> > * User manually initializes the combo using Enchanted Riposte (if fill is selected)
> > * Embolden and Manafication are off cooldown (If option is selected)
> > * Black and White mana reach the overcap protection value.
> 
> First Req: I believe I have your first requirement easy peasy in PR #179 I'd like you to check. Second Req: Part of me also thinks, that last bit I changed, can be expanded for quick check for Embolden & Manafication buffs, but then the description needs to be changed to say that Riposte or both buffs. Third Req: As a separate option, I would think this would be best bled off with Reprise, which would be an option outside of Melee Combo?
> 
> With Also with Second Req I think it would be incompatible with Hold for Double, as it does the holding.

1) Your solution is a lot cleaner. There is a lot of re-used code in the way i wrote this and I wasn't sure how to refactored it.
2) Yeah the description can be updated to specify that if that option is checked, the melee combo will be executed if Mana and Embolden are off cooldown.
3) For the 3rd part, not sure what you mean by bled off. Do you mean if the fill option is selected, the overcap protection should be ignored? The way it currently works is, it will automatically start the melee combo like it does now but instead of it happening at 50 mana, it will happen at whatever the overcap protection is.

Hope that makes sense.

zbee (approved):
Reviewed directly with @nativepapaya:
- the logic appears sound
- adjusted the default mana to overcap at
- features work as they should

@Tartarga I would suggest you re-review, though with the changes after your last comment I think the PR is in a state that works for you, and that you could Approve.

Tartarga:
Sorry. Still juggling adjustments to married life lol. Free time goes straight to playing. 

I'm still not quite fond of (in RDM.cs) all the checks outside / above the Actionfound because it'll run those checks for every rdm action ID.

I'd like to hear feedback on putting the Fill code where I suggested (rdm helper) in my test/reference PR as I think it would be a better spot and not need as many checks

nativepapaya:
> Sorry. Still juggling adjustments to married life lol. Free time goes straight to playing.
> 
> I'm still not quite fond of (in RDM.cs) all the checks outside / above the Actionfound because it'll run those checks for every rdm action ID.
> 
> I'd like to hear feedback on putting the Fill code where I suggested (rdm helper) in my test/reference PR as I think it would be a better spot and not need as many checks

That would be a better spot yeah. Less redundant code.

Tartarga:
I've submitted a PR to your branch. I think I was able to simplify the checks. If it works to your liking, let me know!

nativepapaya:
> I've submitted a PR to your branch. I think I was able to simplify the checks. If it works to your liking, let me know!

Oh thank you, i was trying to figure out how to do it and been having some issues.

zbee:
Closed per most recent comment's most recent edit.

PR #113, opened 2024-11-25
[PVP] Changes and fixes wave 2

edewen:
DNC
ported the dancer partner remind over from pve without the autorotation stuff. added EnAvant option to boost 1 2 combo damage with slider to save charges per user preference.

BLM
Fixed the blizzard side trying to use foul. Not longer a pvp spell. Replaced with xeno as it intends
Fixed missing slider for wreath of ice health threshold.
Added wreath of fire execute mode option with slider. While it goes through guard, it is a lot of dps lost when it can have 50% uptime.  

SAM

MNK
Fixed logic of fire resonance to always use it with phantom rush (largest hit of 12k) unless 2 charges are available and then throw one of a flints or winds reply (2nd largest hits of 8k but aoe)


WrathCombo/Combos/PvP/BLMPVP.cs
```diff
@@ -128,6 +126,8 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                             return Paradox;
                     }
 
+                    if (IsEnabled(CustomComboPreset.BLMPvP_BurstMode_WreathOfIce) && IsOffCooldown(ElementalWeave) && canWeave && PlayerHealthPercentageHp() <= GetOptionValue(Config.BLMPvP_BurstMode_WreathOfIce))
+                        return OriginalHook(ElementalWeave);
```
Kaeris-Tempest, line 130: You need an `Umbral Ice` buff check before you can use Wreath of Ice.

WrathCombo/Combos/PvP/DNCPVP.cs
```diff
@@ -76,12 +83,15 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                             return OriginalHook(CuringWaltz);
 
                         // Fan Dance weave
-                        if (IsOffCooldown(FanDance) && distance < 13) // 2y below max to avoid waste
+                        if (IsOffCooldown(FanDance) && distance < 13 && !enemyGuarded) // 2y below max to avoid waste
                             return OriginalHook(FanDance);
+
+                        if (IsEnabled(CustomComboPreset.DNCPvP_BurstMode_Dash) && ActionReady(EnAvant) && !HasEffect(Buffs.EnAvant) && GetRemainingCharges(EnAvant) > GetOptionValue(Config.DNCPvP_EnAvantCharges))
+                            return EnAvant;
```
Kaeris-Tempest, line 90: I don't think `ActionReady(EnAvant)` is necessary if the charge check is already being handled by `GetRemainingCharges(EnAvant)`.
Automating En Avant seems more harmful than helpful though...

Kaeris-Tempest (commented):
I don't play MNK so I can't help much on that one.

BLM, besides what I pointed out, doesn't have charges on Aetherial Manipulation anymore so that can also be adjusted.

WrathCombo/Combos/PvP/DNCPVP.cs
Kaeris-Tempest: That's fine.

>But hard disagree on en avant being harmful.

Just so we're on the same page, I never said En Avant was harmful; I said _automating_ it was harmful.

Wrath recklessly punting a ranged job out of position will invite trouble more often than than not. If you're facing an enemy (which you will be, because you're attacking it), this feature will yeet you straight into melee range to get cleaved down.

zbee: Probably another case of a feature that could use targeting features; e.g. automatically dashing to exactly where you already are.

edewen: It cant dash to where the player is at. its not targeted. And again I reiterate option. 

I would rather learn to work with it, dashing to where my camera faces,  than have to spam it separately and miss out on all the potency gain as well as the 10% buff for me and my partner because I'm not doing it often enough manually.

Learning to leverage erratic movement of dancer is core to its pvp gameplay. Especially when dancer needs to spend a lot of time in melee to begin with bc of the honing dance 5 yard radius.

Kaeris-Tempest: I don't think anyone is questioning the importance En Avant.

> dashing to where my camera faces

It doesn't do this by default, this is a ReAction feature.

> leverage erratic movement of dancer is core to its pvp gameplay

Of course.
But there is a big difference between planning your dashes _and_ being constantly forced out of position by this.
One match of CC or Frontline with this option on and you'll realize how quickly you hit the floor.

Either way, I'm not against the feature; but as someone who played DNC to crystal I would not touch it.

edewen: As someone who had played dancer for fun and is looking to help the peasant class like myself, I'm fine with disagreeing on it. As long as it is optional. I'm happy.

Issue #41, opened 2024-11-18
[SMN] Option to replace Emerald Ruin III with Ruin III


GayPotatoEmma:
Since in some content (most notably Eureka/BA) your GCD is quite a bit lower (2.12s in my case), using Emerald Ruin III ends up being a damage loss over just using Ruin III it would be nice to have an option to just not use ER3.

After level 72 this would not be relevant as it is then replaced by the higher potency Emerald Rite.

[originally submitted as [XIVSlothCombo #1857](https://github.com/Nik-Potokar/XIVSlothCombo/issues/1857)]

Genesis-Nova:
I'm not really too fussed about adding sub-max level optimizations for niche content.
You're more than welcome to just use Ruin III manually as that is excluded from ability replacements.

PR #121, opened 2024-11-25
[SMN] Add option to replace Emerald Ruin III with Ruin III

GayPotatoEmma:
Figured I'd try my hand at implementing #41 myself, seems to work fine. Any feedback is appreciated.


zbee (changes requested):
Would like to see the restrictions on level and GCD actually put into the code.

GayPotatoEmma:
> Would like to see the restrictions on level and GCD actually put into the code.

I'll have to see if I can figure out how that is done for the GCD, for the level itself it won't actually matter anymore since I've made it so it doesn't override Emerald Rite (the upgraded version from level 72 onwards).

zbee (approved):

Genesis-Nova (changes requested):
The code for it looks fine, but I'd move it down to right above gemshine's place, which is line 733. the `// Gemshine` line

Genesis-Nova (approved):

Taurenkey (changes requested):
This will require a small rewrite.
- Remove checking what Gemshine is, favour checking the gauge for how much attunement is left instead (you're already checking if Garuda attuned so this won't be an issue for other attunements)
- Add a trait check for Ruin Mastery III (ID 476) to ensure it's used before Emerald Rite since at that point it becomes a DPS loss.
- Rewrite the feature description to outline this.

GayPotatoEmma:
Accidentally discarded the commits by trying to resync the fork, sorry about that.

zbee:
@GayPotatoEmma this has become conflicted.

GayPotatoEmma:
> @GayPotatoEmma this has become conflicted.

Will take a look tomorrow

zbee (approved):
Much simpler and cleaner now, further approve.

PR #128, opened 2024-11-26
[MCH] add blazingshot for 92+ AoE

### LINKED ISSUE #126: [MCH] So, the 92+ AoE rotation...
Autocrossbow might just be the worst skill in the game. Once your oGCDs upgrade into Double Check and Checkmate, Autocrossbow now only becomes a gain at 6+ targets. Prior to auto-mode this was out of scope really, but I'd like to raise this again. If nothing else I wouldn't mind a toggle in aoe advanced mode to just never use autocrossbow and substitute blazing shot from level 92 up, since that's applicable to both auto and manual use. 

This also raises the notion in my mind that the AoE threshold setting in the auto mode settings should probably be set per job. For once, this isn't MCH special-casing (though they would benefit, given scattergun), it's that I'm pretty sure tanks want to AoE right from 2 targets, while most other jobs gain on 3. This might be better broken out into a separate issue though?

Kagekazu:
- [x] Cleanup Aoe
- [x] Add blazing shot for 92+ option
  - [x] Add standard to Simple mode
- closes #126

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -2124,6 +2124,10 @@ namespace WrathCombo.Combos
         MCH_AoE_Adv_Hypercharge = 8303,
 
         [ParentCombo(MCH_AoE_AdvancedMode)]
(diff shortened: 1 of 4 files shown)


Kagekazu:
Merged into PR #140

PR #141, opened 2024-11-27
[BLM] Advance mode Astral Fire Tweaks

nativepapaya:
Tightened the timings for Paradox and Firestarter during astral fire as they were a little too conservative. Now Advance mode will try to fit in 4 Fire IVs before every Paradox and Firestarter will not be used unless timing is about to expire before reaching despair.


Taurenkey:
Going to decline this as the current implementation measures against different spell speeds rather than just looking at the element timer, which makes it more flexible overall.

PR #143, opened 2024-11-28
[PVP] Many fixes and changes. All my pvp prs combined. 

edewen:
Sorry for the massive single pr. Removing the vpr code for kaeris to do vpr got all my branches confused and the best way I could do this was to merge all my branches into one, fix vpr conflict, and put in new pr. All the kaeris recommendations from other prs were ported over. 

Untouched They seem to be working fine. WAR, MCH, SAM, PCT, WHM, SGE, SCH

DRK
Fixed the wrong buff code on blackest night, it was overwriting itself. Added was last ability used as good measure bc of server ticks.
Move the Shadowbringer health slider to where it should be in config screen.

PLD
Added Holy Sheltron to the burst combo in melee range/combat.

GNB
Added Heart of Corundum to burst combo with an hp slider to avoid wastage. Can it be used on other? sure, but realistically, GNBS aren't gonna be healing people with it.
Optimized the burst and slimmed down the code significantly. Fixed the 4th combo action.
Added slider for Blasting zone target health %

BRD
Fixed harmonic arrow hijacking the whole combo. Gave it a slider for choice of how many charges to use it at. Added a execute mode for under 12k health so anything above 2 charges should kill your target.

DNC
ported the dancer partner remind over from pve without the autorotation stuff. added EnAvant option to boost 1 2 combo damage with slider to save charges per user preference.

DRG
Fixed Broken 1234 combo adding drakebane. Added execute mode for chaotic spring. It bypasses guard so it is a good option to kill anything under 8k health.

MNK
Fixed logic of fire resonance to always use it with phantom rush (largest hit of 12k) unless 2 charges are available and then throw one of a flints or winds reply (2nd largest hits of 8k but aoe)

RPR
Added Harvest moon execute range at the bottom for enemies under 12k

NIN
Made Mudra Automatic mode an actual option to fix issue with discord report. Added a mudra pooling mode for bunshin burst window. Added the lb execute onto aoe mode as well. 

VPR
I put it back to the original code so that it shouldnt conflict with Kaeris vpr

RDM
Added option to hold charges of corps and displacement if the player chooses.
Fixed displacement simply not working for me. Adjusted to more action readys instead of the getcooldown code that wasnt working for displacement
Tweaked so that the displacement, corps, and embolden buffs all line up with scorch for max damage

SMN
Removed unused fester slider from the config that would confuse the masses
Removed the Canweave from the dps options (necrotize and lb stuff) to streamline things

BLM
Fixed the blizzard side trying to use foul. Not longer a pvp spell. Replaced with xeno as it intends
Fixed missing slider for wreath of ice health threshold.
Added wreath of fire execute mode option with slider. While it goes through guard, it is a lot of dps lost when it can have 50% uptime.
Fixed flare star to only show on the fire button if in astral fire, or the blizzard button if in umbral ice. Allows you to LB and use freeze or flare to change elements and use the star you choose.

AST
Fixed the gravity options to work properly. It was ONLY working with double gravity.
Fixed macrocosmos to work without double gravity if not selected but still hold for double gravity if selected.
Removed pvpimmune tag from lady of crowns as it is an aoe heal.


Nik-Potokar:
@zbee This is good to go now. I have checked and tested if everything functions as intended.

PR #149, opened 2024-11-30
[CORE] cleanup CustomComboPreset.cs

Kagekazu:
- [x] Removed ordering
- [x] Rename all simple modes for uniformity
- [ ] Rename all Advanced modes for uniformity


Kagekazu:
goddammit Ecommons

PR #153, opened 2024-11-30
[Core] Added GetTargetHPPercent Limits

Tartarga:
Clamp down return value range from 0f to 100f
Fixes reported SGE Discord issue


zbee (approved):
Simple fix. LGTM.

Kagekazu (approved):

PR #154, opened 2024-12-01
[PVP] Global Emergency Guard Anti Mash edit

edewen:
Edited the Anti mash feature. 

Original feature prevents you from unguarding by accident in pvp using your combo buttons. It only leaves guard untouched so you can use that to get out of guard. 

The change

If you have the mp AND could use the 15k healing with wasting it, it changes that guard button to recuperate so that can be used to take you out of guard. Exact same functionality, removing guard, but allows you to do it with the heal instead of having to hit guard and then heal resulting in a delay. 

The original request in discord wanted other skills, but to do that, would require unlocking those skills under guard and would likely result in the anti-mash feature having potential fails.

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -5312,6 +5312,10 @@ namespace WrathCombo.Combos
         [CustomComboInfo("Prevent Mash Cancelling Feature", "Stops you cancelling your guard if you're pressing buttons quickly.", ADV.JobID, 3)]
         PvP_MashCancel = 1100030,
 
+        [ParentCombo(PvP_MashCancel)]
+        [CustomComboInfo("Recuperate Option", "Allows you to cancel your guard with Recuperate on the Guard button if health is low enough to not waste it.", ADV.JobID)]
+        PvP_MashCancelRecup = 1100031,
+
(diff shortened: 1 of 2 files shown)


Kaeris-Tempest:
This will cause Guard to break immediately if you're mashing and missing health.

zbee:
If you're mashing guard, yes it would, but not if you entered Guard via the Emergency Guard Feature (i.e. pressing your main combo).

Not sure what else it could or should go on.

I think the answer would be to make this a setting, with a slider of how much time left on the Guard buff is required before triggering the feature (e.g. make the buff go on for at least 2s before letting you exit via Recuperate - e.g. `GetBuffRemainingTime(Buffs.Guard) < 4` would make it require Guard be on for a full second before replacing with recup)

edewen:
get buff is too slow, i still see a flash of recup before it goes back to guard which COULD be bad for fast button spammers, though unlikely. Went with !justused guard and a 2f to keep it from showing up too quickly.

zbee (approved):
Works, without immediately trying to end Guard.
Would still rather this be a specific feature you could opt out of though, but I don't PVP enough to know whether that should be a blocker here.

edewen:
Optional complete and tested.

PR #157, opened 2024-12-01
[PVP] RPR Fixes and updates

edewen:
- [x] Remove double sliders for arcane crest and immortal sacrifice
- [x] Removed plentiful opener, as it is not needed to start lb gen anymore
- [x] Tweaked death warrant to account for the plentiful opener removal. It will fire as long as you are more than 2 stacks below threshold to hold for burst. 
- [x] Fix Harvest moon execute not checking for charges
- [x] Added grim swathe as an option instead of being forced
- [x] Fixed death warrant happening on cd without any options selected as well as the logic of respecting the pooling option
- [x] Clean up unnecessary bools

image

### DIFF
--- a/WrathCombo/Combos/PvP/RPRPVP.cs
+++ b/WrathCombo/Combos/PvP/RPRPVP.cs
@@ -156,7 +156,8 @@ namespace WrathCombo.Combos.PvP
                                     return GrimSwathe;
                             }
                             // Harvest Moon Execute 
-                            if (IsEnabled(CustomComboPreset.RPRPvP_Burst_RangedHarvest) && EnemyHealthCurrentHp() < 12000)
+                            if (IsEnabled(CustomComboPreset.RPRPvP_Burst_RangedHarvest) && GetRemainingCharges(HarvestMoon) > 0 &&
+                                EnemyHealthCurrentHp() < 12000)
                                 return HarvestMoon;
                         }
                     }
--- a/WrathCombo/Combos/PvP/RPRPVP.cs
+++ b/WrathCombo/Combos/PvP/RPRPVP.cs
@@ -137,7 +137,7 @@ namespace WrathCombo.Combos.PvP
                             if (canWeave)
                             {                               
                                 // Death Warrant without pooling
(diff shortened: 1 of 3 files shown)


WrathCombo/Combos/PvP/RPRPVP.cs
```diff
@@ -127,36 +113,37 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
...
-                                deathWarrantReady && distance <= 25 &&
-                                ((plentifulCD > 20 && immortalStacks < immortalThreshold) ||
-                                (plentifulReady && immortalStacks >= immortalThreshold)) || HasEffect(Buffs.DeathWarrant) && GetBuffRemainingTime(Buffs.DeathWarrant) <= 3)
-                                return OriginalHook(DeathWarrant);
+                            // Pooling Plentiful with Death warrant
+                            if (IsEnabled(CustomComboPreset.RPRPvP_Burst_ImmortalPooling))
+                            {
+                                if (IsEnabled(CustomComboPreset.RPRPvP_Burst_DeathWarrant) && deathWarrantReady && distance <= 25 &&
+                                (GetCooldownRemainingTime(PlentifulHarvest) > 20 ||     //if plentiful will be back for the next death warrant
+                                (plentifulReady && immortalStacks >= immortalThreshold) || // if plentiful is ready for this death warrant and you have the charges you want
```
zbee, line 124: Since this is a `>=` instead of an `>`, it seems to me like this would push users under their set thresholds.

zbee (approved):
Seems to function without actual combat testing, builds, and reads like it works.
Only one note, which I don't think should block this merge.

WrathCombo/Combos/PvP/RPRPVP.cs
edewen: The set that they want to use it at 5, then 5 needs to be an option. if there was no =, it would only use it at 6 stacks.

zbee: I believe it is referring to this pictured option, which is the # to save, not the # at which to use stacks.

I read this as "set this to a number we should always stay over", which would be a `>` sign.

image

zbee: I think a wording tweak would be in order to keep the `=`.\
Or a lower maximum for the range, to switch to `>`.

Neither really affects anything, it's just my opinion, but the latter is closer to breaking user settings so maybe just some sort of tweaked wording; it's still an approval for me with or without this change - feel free to mark this as resolved.

PR #158, opened 2024-12-01
Fix PCT SimpleMode + Add Preset confilicts

Nik-Potokar:
Fixed Discord issue "PCT Simple Mode Single Target Issue" where it woudn't know what opener to use so i've added a default one when Simple ST mode is enabled.


zbee (approved):
Simple fix, LGTM.

Kagekazu (approved):

PR #161, opened 2024-12-01
[BRD] Fixes

edewen:
- [x] Fixed level check on dot application logic in st, aoe, and ironjaws
- [x] Added bloodletter logic to aoe simple/adv  ogcd below lvl 45 when you get rain of death
- [x] added logic to songs to allow automatic to play songs in 3 3 12 without a target for bosses  with downtimes

### DIFF
--- a/WrathCombo/Combos/PvE/BRD/BRD.cs
+++ b/WrathCombo/Combos/PvE/BRD/BRD.cs
@@ -533,6 +533,8 @@ namespace WrathCombo.Combos.PvE
                             else if (rainOfDeathCharges > 0) //Dont pool when not enabled
                                 return OriginalHook(RainOfDeath);
                         }
+                        if (!LevelChecked(RainOfDeath) && !(WasLastAction(Bloodletter) && GetRemainingCharges(Bloodletter) > 0))
+                            return OriginalHook(Bloodletter);
                     }
 
                     #endregion
@@ -1083,6 +1085,9 @@ namespace WrathCombo.Combos.PvE
                             else if (rainOfDeathCharges > 0)
                                 return OriginalHook(RainOfDeath);
                         }
+
+                        if (!LevelChecked(RainOfDeath) && !(WasLastAction(Bloodletter) && GetRemainingCharges(Bloodletter) > 0))
+                            return OriginalHook(Bloodletter);
 
                         // Self care section for healing and debuff removal
 
--- a/WrathCombo/Combos/PvE/BRD/BRD.cs
+++ b/WrathCombo/Combos/PvE/BRD/BRD.cs
@@ -150,9 +150,9 @@ namespace WrathCombo.Combos.PvE
                     // At least Lv56 (Iron Jaws) from here on...
 
                     // DoT application takes priority, as Iron Jaws always cuts ticks
-                    if (blue is null)
+                    if (blue is null && LevelChecked(Windbite))
                         return OriginalHook(Windbite);
-                    if (purple is null)
+                    if (purple is null && LevelChecked(VenomousBite))
                         return OriginalHook(VenomousBite);
 
                     // DoT refresh over Apex Option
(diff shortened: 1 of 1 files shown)


WrathCombo/Combos/PvE/BRD/BRD.cs
```diff
@@ -654,9 +662,9 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
                             if (ActionReady(EmpyrealArrow) && JustUsed(WanderersMinuet)) // Used to ensure Empyreal arrow goes off as soon as possible in opener
                                 return EmpyrealArrow;
 
-                            if (canWeave)
+                            if (canWeave || !hasTarget)
                             {
-                                if (songNone)
+                                if (songNone && InCombat())
```
zbee, line 667: If I'm in combat and un-target the boss, this looks to me like it would let me clip my GCDs if I have auto-target-on-attack turned on.

I would probably change this to be `if (canWeave && InCombat())`, for one, but since I want it to go whether I have a target or not ... this check might be superfluous, difficult to say from just what I see here.

WrathCombo/Combos/PvE/BRD/BRD.cs
```diff
@@ -419,14 +422,14 @@ protected override uint Invoke(uint actionID, uint lastComboMove, float comboTim
...
-                                        if (ActionReady(EmpyrealArrow))
+                                        if (ActionReady(EmpyrealArrow) && hasTarget)
                                             return EmpyrealArrow;
                                         return ArmysPaeon;
                                     }
                                 }
                             }
 
-                            if (songArmy && canWeaveDelayed)
+                            if (songArmy && (canWeaveDelayed || !hasTarget))
```
zbee, line 432: Similar to my comment below, I don't think that target checking belongs against a `canWeave` check - this also seems like it would let me clip GCDs if I un-targeted the boss.

WrathCombo/Combos/PvE/BRD/BRD.cs
zbee: Similar issues at :432, :1184, :1219, at least.

zbee (approved):
This is primarily good, though I think it's worth mentioning some potential issues with logic regarding the downtime song maintenance.

WrathCombo/Combos/PvE/BRD/BRD.cs
edewen: That's not a potential problem. Thats literally the point. 

It is so that your song can run when there is no boss to target. with no boss there is no gcd. with no gcd no weaving. no way to automatically run the songs. 

It is coded so that ONLY when in combat, and not targeting something, will the song work on a gcd. once targetting something, it is back on weave only.

zbee: I do understand the point of this, I'm just saying there's more ways to lose a target than just actual downtime: I can arbitrarily un-target at any time, and if it's at song time I would clip.
I was saying that your target checks don't need to be *against* weave checks, it just adds the opportunity to clip, you get the same downtime improvements if you move the target check to happen after and separately from the weave check.

edewen: If the target check is after the weave check. The it would still need to weave. That breaks the function I'm trying to give it. Running songs without something to target. The function of this NEEDS to be on the gcd not weaving when no target is available. 

There are plenty of bosses that step away for a time frame. Waiting on their adds to spawn or doing their super attack. Cagnazzo, zodiark, nidhogg to name a few. This is an actual issue for bard as the song timing is crucial to keep regardless of what the fight is doing. 

I know you are just trying to find the holes in it to make it better. The tiny risk of a gcd clip in the exact moment of a song switchover and someone randomly untargets the boss is far better than songs being out of rotation, which would then delay buff windows and throw off the whole fight for the bard and therefore the party.

zbee: > The function of this NEEDS to be on the gcd not weaving when no target is available.

I see now, yes, I misunderstood the implicit logic there.

> The tiny risk of a gcd clip in the exact moment of a song switchover and someone randomly untargets the boss

Fair point.

Kagekazu (approved):

PR #162, opened 2024-12-01
[PVP] [BRD] Tweaks

edewen:
- [x] Set Silence shot to not happen right after apex. This is to tighten the burst and silence after the larger hits or burst shot or harmonic
- [x] Added line to prevent silence from happening when you already have repertoire buff from repelling shot
- [x] Updated Harmonic execute to happen in stages for each charge level 1-3
- [x] Removed unnecessary line to use pitch perfect at the the end, it replaces the main action id anyway
- [x] Wardens paeon automatic for soft ccs

### DIFF
--- a/WrathCombo/Combos/PvP/BRDPVP.cs
+++ b/WrathCombo/Combos/PvP/BRDPVP.cs
@@ -1,3 +1,4 @@
+using WrathCombo.Core;
 using WrathCombo.CustomComboNS;
 
 namespace WrathCombo.Combos.PvP
@@ -28,6 +29,18 @@ namespace WrathCombo.Combos.PvP
                 EncoreofLightReady = 4312,
                 FrontlineMarch = 3139;
         }
+        internal class Debuffs
+        {
+            public const ushort
+                Silence = 1347,
+                Bind = 1345,
+                Stun = 1343,
+                HalfAsleep = 3022,
+                Sleep = 1348,
+                DeepFreeze = 3219,
+                Heavy = 1344,
+                Unguarded = 3021;
+        }
 
         public static class Config
         {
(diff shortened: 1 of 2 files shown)


WrathCombo/Combos/PvP/BRDPVP.cs
```diff
@@ -41,47 +54,54 @@ internal class BRDPvP_BurstMode : CustomCombo
...
             {
 
                 if (actionID == PowerfulShot)
                 {
                     var canWeave = CanWeave(actionID, 0.5);
+                    uint harmonicCharges = GetRemainingCharges(HarmonicArrow);
 
                     if (!PvPCommon.IsImmuneToDamage())
                     {
+                        if (IsEnabled(CustomComboPreset.BRDPvP_Wardens) && InPvP() &&  //Autowardens set up only for soft ccs, it cant be used while cced like purify
```
zbee, line 67: I could be mistaken, but I think this `InPVP()` check is completely unnecessary inside of PVP combos.
If it's not, that should probably be changed at some point 😅

zbee (approved):
Simple changes, LGTM; one unimportant note.

WrathCombo/Combos/PvP/BRDPVP.cs
edewen: I would have to check. I belive i was assuming that inpvp check ensues you are actually in pvp combat actively?

Taurenkey: It does not check that, only if you're in a pvp zone.

zbee: The check is not done. It is not done because the combo doesn't even run unless a PVP-only ability is being used, `Powerful Shot`.
So this check doesn't provide anything.
Not to say that it is a real problem either, still an approval from me, but it'd be best removed.

PR #163, opened 2024-12-01
[PVP] Purify WHM CC

edewen:
Added WHM cc, Miracle of nature, to the options for purify using the checkboxes. With great struggle and a lot of help from zbee.


zbee (approved):
Simple addition, and already had the chance to test it; works well, and will be slightly easier to add new options here in the future.

PR #164, opened 2024-12-01
[PVP] Nin Seiton Unsealed fix

edewen:
Added SeitonUnsealed Buff to the code for Seiton Tenchu so it can do the followup LB attack.


zbee (approved):
Logic is sound.

PR #184, opened 2024-12-10
[DNC] [MCH] Add `has hostile target` checks to prepull features

zbee:
- [X] Adds check for a hostile target to all DNC prepull features
- [X] Adds same check to MCH ST openers and prepull `Reassemble`s


Kagekazu (approved):
looks good!

Taurenkey (commented):
See my comment [here](https://github.com/PunishXIV/WrathCombo/pull/185#discussion_r1880657344)

PR #186, opened 2024-12-11
[SMN] Fixed swiftcast egi feature.

Genesis-Nova:
Moved the last bit of the swiftcast egi part into the correct code block.


zbee (approved):
Simple fix as discussed in dev chat, LGTM.

Kagekazu (approved):

PR #190, opened 2024-12-12
[Core] Account more widely for player hitbox

zbee:
- [X] Moves the accounting for the enemy hitbox from `InMeleeRange` to `GetTargetDistance`, as suggested by Tauren, and with testing with spells not just melee


zbee:
The addition accounting for the hitbox in `GetTargetDistance` in db0955c will break `InMeleeRange`, this would still need merged.

Taurenkey (approved):

PR #196, opened 2024-12-14
[Core] Make chat message prefixes more consistent

zbee:
- [X] Switch chat messages (except MOTD) over to use EC's Duolog, for `[WrathCombo]` prefix consistency


Taurenkey (approved):
Funnily enough, this was on my todo list aswell.

PR #201, opened 2024-12-15
[Various] Reorder presets for my classes

Kagekazu:
- [x] Reorder BLM
- [x] Reorder DRG
- [x] Reorder MCH
- [x] Reorder MNK
- [x] Reorder RPR
- [x] Reorder SAM
- [x] Reorder VPR


zbee (approved):
Slightly concerned about the removal of checks of things like `CustomComboPreset.VPR_ST_CDs`, since its sub-options could still be enabled, as far as I am aware.

But I assume that was tested for and otherwise LGTM.

Kagekazu:
> Slightly concerned about the removal of checks of things like `CustomComboPreset.VPR_ST_CDs`, since its sub-options could still be enabled, as far as I am aware.
> 
> But I assume that was tested for or is not the case, and otherwise LGTM.

Yeah, can always re add it should it be an issue.

PR #207, opened 2024-12-16
[MNK] few fixes

Kagekazu:
- Add Formless Fist to AoE
- Add better Meditation usage for ST and AoE
- Fix Formless Fist bug when not in combat but in PB
- General cleanup

requires merge #201


zbee (approved):
The logic changes look sound, as do the meditation changes.

PR #217, opened 2024-12-20
[Docs] Contributing Guide

zbee:
Adds a basic contributing guide I can expand upon more as I go.
Primarily, this is the sort of information I wish I had when I began contributing, and that will be the general guiding force for what `/docs` I add.

- [X] Adds a contributing guide, with the templating information that was previously kept in the private resources gist.
  - [X] Includes basic snippets of documentation about the main bits of the code.
  - [X] Includes all designed templates for jobs.
  - [X] Will be linked to when creating a PR (probably for non team members?)
  - [X] Establishes somewhere to link to all guides from, which would be under `/docs`
  - [X] Acts as a quick-linkable resource for any contribution questions (especially because of the above)


Kagekazu (approved):
looks good

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+- Advanced Mode - AoE
+- `<combo name>` Feature
+  - `<option name>` Option
+
+### [Healers] Healing Feature Naming Template
+- Simple Heals - Single Target
+- Simple Heals - AoE
+
+### [Tanks] Mitigations Option template:
+> NOTE: the Mitigation option must be called "Mitigation Options" in order to get highlighted.
```
Taurenkey, line 36: This can be removed now

PR #218, opened 2024-12-20
[BRD] Fixes and Tweak

edewen:
- Fixed one sided conflict for Heavy shot to straight shot feature

- Reordered the CCP for bard now that Kages removal of number system is done

- Extended the aoe no waste limit up to 50% per conversation in disc suggestion


zbee (approved):
Simple enough, LGTM - Thanks for taking care of that one-way.

Kagekazu (approved):

PR #220, opened 2024-12-20
[PvP] BLM Wrath Update

Kaeris-Tempest:
Black Mage has been revisited for Patch 7.15.

## Major Changes
- Complete logic rework.
- Fire and Blizzard spells have been condensed into one button.
- Smart Paradox usage for higher damage and fire-aspected uptime.
- Elemental Weave now accommodates both Fire and Ice wreaths.
- Several actions now have toggles to include defensive usage.

## Minor Changes
- Lethargy will no longer be used against NPCs.
- Xenoglossy prefers saving charges when used against NPCs.
- Aetherial Manipulation removed as it no longer grants Swiftcast.

## Other Changes
- Removed all weaving checks and obsolete IDs.
- Added missing actionIDs for Fire/Blizzard III and High Fire/Blizzard II.
- Added `Aetherial Manipulation Feature` for quick debuff cleansing.

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -6415,64 +6415,41 @@ public enum CustomComboPreset
     #region BLACK MAGE
 
     [PvPCustomCombo]
-    [CustomComboInfo("Burst Mode", "Turns Fire and Blizzard into all-in-one damage buttons.", BLM.JobID)]
+    [ReplaceSkill(BLMPvP.Fire)]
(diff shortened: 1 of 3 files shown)


Kaeris-Tempest:
> edited my comment above - ill play it some more and let u know if i find anything else.

Thanks.

Regarding the "staying on ice" part, this only happens when you reach `Umbral Ice III` as it wants to spend `High Blizzard II` even if you're not moving, before going back to fire - otherwise you're giving up a free 9,000 damage (+4,500 AoE) instant spell; but it does mean delaying going into fire for one GCD. This does not apply during Soul Resonance.

The movement aspect I'm still a bit undecided on, mostly because it serves as a "control stick" for which element you want to be in. For example, if you just want to finish off a low health target, start moving to use a quick ice spell; instead of hard casting fire spell.

It takes a bit of getting used to, but I feel the tight controls are better overall, but I'm open to re-evaluating it over time.

PR #231, opened 2024-12-23
[Various] Update classes

Kagekazu:
- update all my classes (BLM, DRG, MCH, MNK, RPR, SAM, VPR) to be uniform
- remove helper in helper
- reorder customcombo


zbee (approved):
All looks sound.

PR #238, opened 2024-12-24
[WAR] Fix Action for Inner Release

zbee:
- [X] Changes `Inner Release` combo to go onto `Inner Release` or `Berserk` instead of only onto `Berserk`


Kagekazu (approved):
LGTM

PR #246, opened 2024-12-27
[BLM PVP] Movement Fix for requiring astral fire

edewen:
2 Bugs reported. 

1. When starting out, it forces you fire first even if you are trying to move. Ignoring the movement check waiting for astral fire. 

2. When moving, for some reason, when getting to umbral 3, instead of casting ice spells while moving, it was trying to cast fire spells and failing to operate. 

Removing this one tag, seems to have fixed both issues for me in testing.

### DIFF
--- a/WrathCombo/Combos/PvP/BLMPVP.cs
+++ b/WrathCombo/Combos/PvP/BLMPVP.cs
@@ -171,7 +171,7 @@ namespace WrathCombo.Combos.PvP
 
 
                     // Basic Spells
-                    return isMovingAdjusted && Config.BLMPVP_BurstButtonOption == 0 && hasAstralFire
+                    return isMovingAdjusted && Config.BLMPVP_BurstButtonOption == 0
                         ? OriginalHook(Blizzard)
                         : OriginalHook(actionID);
 


Kaeris-Tempest:
>     1. When starting out, it forces you fire first even if you are trying to move. Ignoring the movement check waiting for astral fire.

This is caused by a change that was made over my original implementation.
https://github.com/PunishXIV/WrathCombo/commit/4af5e7c638421e6d100a7e40cef93800c7cb591c#diff-f24c0386dd96d2de354b131e4163ad88b214cc99affbc8ab00e9c62c8c6aaa58L170

>     2. When moving, for some reason, when getting to umbral 3, instead of casting ice spells while moving, it was trying to cast fire spells and failing to operate.

The movement code doesn't care about Umbral Ice buffs.

> Removing this one tag, seems to have fixed both issues for me in testing.

This will just cause the delay to be applied globally, even outside Astral Fire (you don't want that).

edewen:
Looking at the issue further, it is not just ice 3. It is any time there is no astral fire. 

I thought it was ice 3 because what happens is when you are in fire and start moving, it will use blizzard first, which puts you into ice. Once it is in ice, it will paradox putting to ice 3, but the movement check is failing with any umbral ice buff because you dont have astral fire. It may not care about umbral ice buffs, but it seems to want to care about having fire. Which is wrong. 

The same issue, of caring about fire, is causing, when having neither buff, to be forced to start with a fire spell. 

When moving for more than the set time frame, it will use ice spells not fire. There is no delay that is touched by removing the astral fire check to cast a blizzard spell. This change fixes the code currently to work as it should.

zbee (approved):
Makes good sense with the recent explanation: just more fully accounting for the 2-button change, it would seem.

Kaeris-Tempest:
To reiterate, the code was never meant to be the way it is currently.

I added the movement delay specifically to avoid overwriting Astral Fire as easily when moving. It worked flawlessly when both `isMoving` and `isMovingAdjusted` were part of the filtering operator.

However, with the removal of `hold mode` and how the movement conditionals were rewired, the Astral Fire check is now functionally useless and is most likely just causing problems.

I don't know how the logic operates currently, so I guess keep removing things until it behaves.

edewen:
I get the this wasn't your vision of it and that things were changed after you put out your pr. And frankly, I don't care about the blame game or credit. It serves no purpose at this point and really doesn't matter. It was put out as a wrath update. That's the name of the team, not a member. The feature just needs to work, which currently it isn't. That should be what matters. 

You've expressed you have zero interest in fixing this issue. I do. So I tested and found that you are right and the astral fire check is indeed causing the issue. By removing it, the bandaid is applied until a more robust solution could be added beyond the slider allowing the player to notate how long to be moving before it changes to Ice.

PR #247, opened 2024-12-27
[VPR] Reawaken Legacy option

Kagekazu:
- Remove conflict with ST
- Put on Reaving Fangs instead of Steel Fangs
- Make low lvl/no opener behave again.


zbee (approved):
Conflicts look to all go both ways, LGTM.

PR #251, opened 2024-12-28
[BRD] Buff Realignment

edewen:
Changed the order of buffs in the bard code to match the balance.

Was
Lateweave radiant into BV/raging. 

Now
Lateweave BV int radiant/raging.

This will not improve numbers. But it will make it work better with the hard coded opener so my code doesn't have a gcd of drift on the 2 min fixing its buff order to get the extra gcd of 5% damage instead of increased direct hit chance. 

Added quick edit. 
Raging jaws was trying to happen without dots on a target bc the debuff check of being less that 35 was still coming back as a yes when it was not present. Weird interaction with chaotic switching targets. Simple moved it below the dot application so if something doesn't have dots it wont even consider raging strikes ironjaws refresh. It will apply dots.


zbee (approved):
LGTM, good to align with The Balance on this, even if somewhat unimportant (besides addressing the drift this was implemented for, obviously).

PR #255, opened 2024-12-29
[NIN PVP] Unsealed Seiton Tenchu Health Check

edewen:
Shifted the Health check slider for seiton tenchu outside of the or statement so it also applies to unsealed Seiton techu


zbee (approved):
LGTM as a fix for this [discord issue](https://discord.com/channels/1001823907193552978/1323014770789580902)

PR #257, opened 2024-12-30
[MCH] Queen and Excavator

Kagekazu:
- Add boss logic for when to use what part of queen
- Add boss logic for when to use excavator


zbee (approved):
Simple stuff, LGTM.

Would like to see these boss checks as an option added to the Advanced Modes though; a la DRK:

Taurenkey (changes requested):
Going to agree with @zbee, would like to see options added for boss checks rather than being enforced for advanced modes (you can enforce it without an option in simple, that's not an issue)

Kagekazu:
will re open once i find a way im happy with that doesnt lock queen or excavator to boss only content as option.

PR #258, opened 2024-12-31
[Core] Make combo action checks consistent

zbee:
> Apologies for doing this after the codeowner addition, but it needed to be after a full merge.

The point of this endeavor was to ensure that all combos had action checks **before** logic started being applied, which is a drain on FPS.
While doing this, I also made all of the action checks consistent with each other in the following ways:
- Inverted any that had nesting
- Used "is" verbage for all of them, since that is what was dominant, and it is the same IL code in these cases

In doing this, <ins>**11**</ins> combos were fixed to no longer do any substantial logic before checking the action, <ins>**2**</ins> of which were particularly egregious
Just for reference, these combos were across:
- DNC
- MCH
- PCT
- RDM
- RPR
- SGE !
- SMN !

Most of these changes were very much a simple inversion, only RDM, SMN, and MCH need to be double-checked.

### DIFF
--- a/WrathCombo/Combos/PvE/AST/AST.cs
+++ b/WrathCombo/Combos/PvE/AST/AST.cs
@@ -273,76 +273,76 @@ namespace WrathCombo.Combos.PvE
             protected internal override CustomComboPreset Preset { get; } = CustomComboPreset.AST_AOE_DPS;
             protected override uint Invoke(uint actionID)
             {
-                if (GravityList.Contains(actionID))
-                {
-                    //Variant stuff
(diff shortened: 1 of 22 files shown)


Kagekazu:
BLM - DRG - MCH - MNK- RPR - SAM - VPR seem to be good.

MCH might conflict with my PR, but ill fix that if needed.

Kagekazu (approved):
BLM - DRG - MCH - MNK - RPR - SAM - VPR seem to be good.

MCH might conflict with my PR, but ill fix that if needed.

PR #261, opened 2024-12-31
[BRD] Fix the raging jaws for simple too

edewen:
Same fix as last pr but to simple too. putting dot reapplication above raging jaws logic. simple shift of text.


zbee (approved):
:)

Kagekazu (approved):

PR #262, opened 2024-12-31
Weaving guardrails (Obsolete use check)

Tartarga:
Added some guardrails incase out of habit Weave(ActionID) happens


zbee (approved):
Thanks so much for adding this

PR #282, opened 2025-01-13
[PCT] Cleanup + Bosscheck

Kagekazu:
- Cleanup of nested `if`'s
- Add Bosscheck for Starry Muse on ST ( Can add to AoE if needed/wanted )
- Add default values to all options
- Cleanup Namespace


Kagekazu:
Need to check other functions that use starry sky cooldown to work outside of bosses aswell

PR #286, opened 2025-01-16
[MNK] Dragon Kick added to Solar Lunar Opener

tak-st:
I am submitting this pull request because an element is missing, probably due to a mistake.

If this opener is intended to start with a perfect balance, then this change is not necessary.


zbee (approved):
Every Balance opener I see begins with `Dragon Kick`.

Kagekazu (approved):
Probably a copy paste error on my end. 

It should have Dragon kick yes. 

thank you!

PR #297, opened 2025-01-20
[Multiple] Buffchecks + Cleanup

### LINKED ISSUE #296: DRG action change bug same as MNK
After updating to v1.0.0.9 when playing DRG or MNK with or without autorotation, any button that has an action change setting gets stuck, as the rotation keeps trying to use the most recent action change button whilst its on CD for e.g. DRG gets stuck on Geirskogul and wont use anything other than the normal GCD's until its back off of CD. Monk does the same but with Riddle of Wind. at the moment it seems to work if I turn off all of the action change buttons, but I dont think this is how its supposed to be is it?

I was using it last night and had no issues.

Thanks for taking your time to read.
N.R.

Kagekazu:
- Add better buff checks for DRG, MCH, SAM, MNK
- Fixes #296 
- Cleanup of all my classes ( BLM, DRG, MCH, MNK, RPR, SAM, VPR )

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -2586,7 +2586,7 @@ public enum CustomComboPreset
     MCH_ST_AdvancedMode = 8100,
 
     [ParentCombo(MCH_ST_AdvancedMode)]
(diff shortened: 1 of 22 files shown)


Kagekazu:
going to split it per class for easy overview

PR #314, opened 2025-01-25
[DNC] Buffcheck fix

edewen:
Oddly aoe only, no idea why. After technical finish it was locking on tech step as long as Tiliana was usable through the flourishing finish buff.

Buffcheck issue from techstep turning into Tiliana Added normal fix to the needtotech line. Fixes issue but doesnt explain why it wasn't happening to st, added there just to be safe though.


zbee (approved):
Thanks for hopping in and checking that for me 👍🏻

PR #316, opened 2025-01-26
[Tanks] Add interrupt into the combos, [Core] Rework settings tab

zbee:
- [X] Adds Interrupting to all tanks' combos
      ST: added option to interject if the target's cast is interruptible
      ST Simple: interrupts if the target's cast is interruptible
      AoE: added options to interject if the target's cast is interruptible, or to stun if the target is casting
      AoE Simple: interrupts if the target's cast is interruptible, and stuns if the target is casting
  - [ ] DRK (handled in #320 to avoid conflicts)
  - [X] GNB
  - [X] PLD (Included `Shield Bash` in the Stun option, in addition to `Low Blow`)
  - [X] WAR
- [X] Adds new Setting for Interrupt Delay.
      Which delays interrupts by X% of the total cast.
      If a value is not passed to `CanInterruptEnemy()` then it will use this setting, and only return `true` if the cast is over that percentage done.
- [X] Reworked the settings tab. Was getting cramped.
      new settings tab

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -1333,6 +1333,8 @@ public enum CustomComboPreset
 
     #region DARK KNIGHT
 
+    // Last value = 5055
+
     #region Simple Mode
 
     //TODO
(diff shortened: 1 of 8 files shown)


zbee:
@Akechi-kun would you be able to give this a test perhaps?

PR #335, opened 2025-02-07
[All] Use `All.SavageBlade` over `11`

zbee:
- [X] Just switch uses of `Savage Blade` to call a constant,\
      to make it easier to spot and easier to track down all instances of this behavior
- [X] Removed the needless `OriginalHook` call around it.\
      It's an action no longer in any job we use when we want to block input, it's not going to Action-Change, and we don't care if it does.


Tartarga (approved):
Looks good to me, didn't realize this was a thing

Random thought for non-pve namespace bits of code, is there a case for doing a "using static"?

zbee:
> [...]is there a case for doing a "using static"?

@Tartarga could you explain the benefits?

PR #354, opened 2025-02-14
[Core] Try to call out `Savage Blade` more

zbee:
- [X] Adds (or updates) every preset/etc that returns `Savage Blade` to specifically call it out


Taurenkey:
Ok but why is my character trying to use Savage Blade, is he dumb?

zbee:
ChatGPT, how do I say yes to that?

Kagekazu (approved):

PR #361, opened 2025-02-15
[SGE][SCH] Optimize AoE Heals code

zbee:
- [X] Optimizes the AoE heals code in a minor way, avoiding fetching the HP for every prioritized ability


PhoenixtheII:
This could actually be applied to SCH too... since it's loop is the same?

Tartarga (approved):

PR #366, opened 2025-02-18
[Healers] Dot update

Kagekazu:
- Add Bosscheck to DoT uptime
- remove hp% slider for DoT
- ran cleanup

### DIFF
--- a/WrathCombo/Combos/PvE/AST/AST.cs
+++ b/WrathCombo/Combos/PvE/AST/AST.cs
@@ -48,7 +48,7 @@ internal static partial class AST
             if (!InCombat())
             {
                 if (IsEnabled(CustomComboPreset.AST_DPS_AutoDraw) &&
-                    ActionReady(OriginalHook(AstralDraw)) && 
+                    ActionReady(OriginalHook(AstralDraw)) &&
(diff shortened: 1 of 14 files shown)


zbee (approved):
This seemed to test fine, and looks fine overall. Approve.

I will say that **I don't** really think this is an exact replacement though, and at least in my opinion isn't an improvement.

I'd rather see an "All enemies / Not Bosses" option _**apply**_ to the hp% slider, rather than a similar option replace it entirely.
With this change, you'd be able to prevent wasting time DoT'ing when you ST the last couple trash mobs, fine; but it would also not let you DoT the single big enemies anymore that you sometimes get in dungeons, like Alexandria, and that would be very unfortunate.

In case I'm being unclear, I mean it would behavior similarly to the Difficulty options in DRK for Delirium: the hp% slider is there, but you can set the Difficulty option to make that slider only apply to easy content - So I can make Delirium stop being used at 10%hp ... but only in dungeons, etc, NOT bosses. Different, but similar.
image

---

But this is all fairly moot if @Taurenkey's addition went in, where DoTs are only done if they'd be effective? I don't think that is in though.

zbee:
Sorry for the late review essentially requesting changes (the lateness is a bit of why it's an approval), but this is when I took the time to look into it is all 😬

zbee:
Pushed a commit that fixes this run-on, and the indentation.
image

zbee (approved):

PR #367, opened 2025-02-19
[PVP] Small fixes

edewen:
Readd removed meteordrive option to monk. Seems it got removed with the update to targetimmune coding. 

Remove glare 4 option from WHM. The game itself replaces glare 3 with 4, thats not something we do not do can we stop it.

Re-added the config sliders for gnb pvp that were mistakenly removed by the large pve mitigation addition back on dec 3rd.


zbee (approved):
LGTM

Kagekazu (approved):
LGTM

PR #369, opened 2025-02-20
[GNB] Fix error + Variables Refactor

Kagekazu:
- Refactor all Variables to Helper class
- Add Bosscheck for NoMercy on ST
- Fix error on Config


zbee (approved):
Works exactly as expected when tested locally in the PR I included it in.

PR #382, opened 2025-02-27
[SGE] Taurochole adjustment

### LINKED ISSUE #378: [SGE] Taurochole to Druochole and level sync
When this option is activated and that the level is synced below 62 (when Taurochole is unlocked), the action is greyed out.

I believe it would be nice to change it to Druochole in this case too as Taurochole is simply not available.

Tartarga:
Taurochole will become Druochole when player level is below Taurochole

Fixes #378


Kagekazu (approved):
LGTM

zbee (approved):

PR #396, opened 2025-03-05
[GNB] Adjusted `Opener` option to support `Lv90` and `Lv100` openers

Akechi-kun:
Openers included should be more than suitable for all content.
As for `<= Lv89`, the logic that's already implemented should be able to easily handle it.

image

- [x] Add new openers for `Lv90` & `Lv100` that support with or without `Lightning Shot` and normal or early `No Mercy`
- [x] Add `VeryDelayedWeaveSteps` into `WrathOpener` & very slight logic adjustment to accommodate for its implementation
_NOTE_: This is for weaving abilities as late as possible. Use-cases involve GNB's `No Mercy` with appropriate skill speed (<= 2.47). 
Let me know if there's a better way to implement this, but so far it works pretty sound from testing.


WrathCombo/Combos/PvE/GNB/GNB.cs
```diff
@@ -177,9 +177,18 @@ protected override uint Invoke(uint actionID)
...
             #region Standard
-            if (IsEnabled(CustomComboPreset.GNB_ST_Advanced_Opener) &&
-                Opener().FullOpener(ref actionID))
-                return actionID;
+            if (IsEnabled(CustomComboPreset.GNB_ST_Advanced_Opener))
+            {
+                if (Config.GNB_Opener_StartChoice == 1)
+                {
+                    if (Opener().OpenerStep == 1)
+                        Opener().OpenerStep++;
```
zbee, line 185: Did this still need changed?

PR #410, opened 2025-03-19
[Core] Variant Inheritance logic fixes

zbee:
- [X] Adds `IsEnabled` check to `Ultimatum`
- [X] Makes Variant action logic consistent in checking abilities are ready

### DIFF
--- a/WrathCombo/Combos/PvE/Content/Variant.cs
+++ b/WrathCombo/Combos/PvE/Content/Variant.cs
@@ -50,22 +50,24 @@ internal static class VariantActions
     }
 
     internal static bool CanRampart(CustomComboPreset preset, WeaveTypes weave = WeaveTypes.None) =>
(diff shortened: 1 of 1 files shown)


Tartarga:
In the car atm, but I remembered there was a difference between IsEnabled(action) and actionready. With the changes to actionready is this a safe change?

zbee:
> With the changes to actionready is this a safe change?

I _added_ `ActionReady` to them all, and _added_ `IsEnabled(action)` to `Ultimatum`, so I believe so.

I can check in-game shortly, and if not we can use `IsOffCooldown` for them all instead, but yeah in their current form there is no check at all for if the abilities are actually ready to be used.

zbee:
Checking `ActionReady` it is does look safe for Variant actions, and looking at the history, the replaced Variant logic mostly uses it as well.
So that should be fine (will still test shortly, and edit if it's not fine).

So that just leaves the addition of the `IsEnabled(action)` then for `Ultimatum`, which is and was used for all the other Variant actions.

PR #414, opened 2025-03-21
[Core] Silenced some warnings w/ inheritance.

Tartarga:
However I am still confused as to why it is complaining about just the Healers


zbee:
The Healer complaints of concern were fixed here: 7a875ac

PR #422, opened 2025-03-24
[PvP] Added RoleActions with dummy IDs

Tartarga:
as for the csproj, sorry, too tired to remember how to undo that. idk how it happened


Tartarga:
Digging through Garland, the player has a buff to what action is equipped.
Sounds like we can check that.
Casters IDs Added

PR #427, opened 2025-03-29
[Core] Removed legacy IsEnabled(uint). ActionReady handles it now.

Tartarga:
@zbee If you don't mind, can you take a quick gander at DRK_Helper, and let me know if you want me to remove IsEnabled, or if you want to do something different.


zbee:
@Tartarga I pushed that removal for you

zbee (approved):

PR #442, opened 2025-03-31
[Healer] Fix DOT threshold

msdx321:
* We don't set the Adv boolean anymore.

### DIFF
--- a/WrathCombo/Combos/PvE/SGE/SGE.cs
+++ b/WrathCombo/Combos/PvE/SGE/SGE.cs
@@ -234,7 +234,7 @@ internal partial class SGE : HealerJob
                             if (TraitLevelChecked(Traits.OffensiveMagicMasteryII))
                                 dotDebuff = Math.Max(dotDebuff, GetDebuffRemainingTime(Debuffs.EukrasianDyskrasia));
 
-                            float refreshTimer = Config.SGE_ST_DPS_EDosis_Adv ? Config.SGE_ST_DPS_EDosisThreshold : 5;
+                            float refreshTimer = (Config.SGE_ST_DPS_EDosisThreshold != 0) ? Config.SGE_ST_DPS_EDosisThreshold : 5;
                             int hpThreshold = Config.SGE_ST_DPS_EDosisSubOption == 1 || !InBossEncounter() ? Config.SGE_ST_DPS_EDosisOption : 0;
 
                             if (dotDebuff <= refreshTimer &&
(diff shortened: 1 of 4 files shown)


Tartarga:
Nice catch!
I do have 2 nitpick concerns on Sage.
1) After talking to Kage, 5 seconds default on single target was settled on due to Eukrasia
2) AoE's DoT should not rely upon ST's slider.  I haven't had anyone ask for a fine tuned slider for it.  If you wish for it to be controlled by a slider, please add a new slider control, else please revert the AoE code

edewen:
is this something that would also affect brd. not sure what it means but we has the dots

Tartarga:
> is this something that would also affect brd. not sure what it means but we has the dots

Shouldn't this is a Healer UI to healer feature issue.

WrathCombo/Combos/PvE/SGE/SGE.cs
```diff
@@ -116,9 +116,11 @@ protected override uint Invoke(uint actionID)
                 float dotDebuff = Math.Max(GetDebuffRemainingTime(currentDosis.DebuffID),
                     GetDebuffRemainingTime(Debuffs.EukrasianDyskrasia));
 
-                const float refreshtimer = 3; //Will revisit if it's really needed....SGE_ST_DPS_EDosis_Adv ? Config.SGE_ST_DPS_EDosisThreshold : 3;
+                float refreshTimer = (Config.SGE_ST_DPS_EDosisThreshold != 0) ? Config.SGE_ST_DPS_EDosisThreshold : 3;
```
Kagekazu, line 119: AoE's DoT should not rely upon ST's slider. I haven't had anyone ask for a fine tuned slider for it. If you wish for it to be controlled by a slider, please add a new slider control, else please revert the AoE code

WrathCombo/Combos/PvE/SGE/SGE.cs
```diff
@@ -234,7 +236,7 @@ protected override uint Invoke(uint actionID)
                             if (TraitLevelChecked(Traits.OffensiveMagicMasteryII))
                                 dotDebuff = Math.Max(dotDebuff, GetDebuffRemainingTime(Debuffs.EukrasianDyskrasia));
 
-                            float refreshTimer = Config.SGE_ST_DPS_EDosis_Adv ? Config.SGE_ST_DPS_EDosisThreshold : 5;
```
Kagekazu, line 237: This was set to 5 to make sure it refreshes before it runs out while having time to do eukrasia + dosis.

Kagekazu (changes requested):

WrathCombo/Combos/PvE/SGE/SGE.cs
msdx321: Sorry I wasn't noticed there was no tick when remaining time is less than 2s. 5s should be default.

msdx321:
I've re-pushed the branch with requested changes.

Kagekazu (approved):
Thanks for making the changes to SGE, looks good to me!

Tartarga (approved):

PR #450, opened 2025-04-02
[PvP] Separated UserOptions from PvE

Tartarga:
Testing to be done, not all user options were even wired up o_o

ToDo: verify ingame


Tartarga:
Closing. Letting edewin perform 90% of these changes with his new action PRs. I don't want to get in their way

PR #452, opened 2025-04-03
[PVP][Casters] PVP role action and move pvp config to pvp  file

edewen:
Added Phantom dart to rdm, smn, and pct
Added phantom dart health slider  to all 4 casters
Moved pvp config from the class config file to the pvp like all the others.


Kagekazu (approved):
LGTM

PR #482, opened 2025-04-13
[WHM] Dot spamming fix

edewen:
Rewrote white mage dot in the bard style copying the purple blue setup given to me a while back

Tested on dummy, 4 man ifrit, and ultima. all 3 dots work, and reapply correctly. hp threshold code remands untouched.


zbee:
I moved that `hpThreshold` code into your method, and moved it.
I did however also revert the change away from using `AeroList`, as I find that a lot cleaner.
Have a look. Can revert it if you like, of course.

edewen:
If it works it works. Whm is your code I'm just here to help out.

PR #483, opened 2025-04-14
[Various] Cleanup after merge

Kagekazu:
- General cleanup after the status merge
- Added better fires reply for MNK
- Add me as sole owner for SAM


zbee (approved):
owner change looks good

PR #490, opened 2025-04-15
[Various] Remove uneeded checks

Kagekazu:
removing lots of unneeded checks since new functions 
rewrote some stuff that needed updating


Tartarga (approved):

zbee (approved):
The removal of unneeded checks looks sound

Issue #176, opened 2024-12-05
[AST] quick targets uses card after the use of divination? (saving them for burst)


Astrouser03:
Hello! I love using quick card targeting but I was wondering if @zbee could possible work on this to bring it to the game ! I wondering if the quick targeting system can be used after the use of divination saving card for burst phase?

zbee:
Closed by #491

PR #491, opened 2025-04-15
[AST] Card pooling options

edewen:
Pooling option and small variant fix I saw was needed. 
#176 and discord request.

- [x] Added card pooling option for single target rotation into divination window
- [x] Added card pooling option for aoe rotation into divination window
- [x] Added lord pooling option for single target rotation into divination window
- [x] Added lord pooling option for aoe rotation into divination window
- [x] fixed single target variant rampart calling use of spirit dart
- [x] moved single target spirit dart up with rampart and outside of requiring combust being enabled

### DIFF
--- a/WrathCombo/Combos/PvE/AST/AST.cs
+++ b/WrathCombo/Combos/PvE/AST/AST.cs
@@ -61,7 +61,7 @@ internal partial class AST : Healer
             if (InCombat())
             {
                 //Variant stuff
-                if (Variant.CanSpiritDart(CustomComboPreset.AST_Variant_Rampart))
+                if (Variant.CanRampart(CustomComboPreset.AST_Variant_Rampart))
                     return Variant.Rampart;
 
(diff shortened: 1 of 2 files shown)


Kagekazu (approved):

zbee (approved):
Clean addition

WrathCombo/Combos/PvE/AST/AST.cs
```diff
@@ -60,6 +62,9 @@ protected override uint Invoke(uint actionID)
             {
                 //Variant stuff
                 if (Variant.CanSpiritDart(CustomComboPreset.AST_Variant_Rampart))
+                    return Variant.Rampart;
```
zbee, line 65: Looks like this is checking `CanSpiritDart` for `Variant.Rampart` still, @edewen.

edewen: Well wtf. Fixed.

PR #493, opened 2025-04-15
[WHM & SGE] Regen and shield fixes

edewen:
The GetStatusEffect was not working and preventing any use of Regen. Rewrote it for GetStatusEffectRemainingTime, simplified yet functions as it should.

Added the missing !s to sage eukrasian diagnosis setup to make the shield checks function correctly again,. 

#501

### DIFF
--- a/WrathCombo/Combos/PvE/SGE/SGE.cs
+++ b/WrathCombo/Combos/PvE/SGE/SGE.cs
@@ -367,12 +367,10 @@ internal partial class SGE : Healer
             }
 
             if (IsEnabled(CustomComboPreset.SGE_ST_Heal_EDiagnosis) && LevelChecked(Eukrasia) &&
-                GetTargetHPPercent(healTarget, Config.SGE_ST_Heal_IncludeShields) <=
-                Config.SGE_ST_Heal_EDiagnosisHP &&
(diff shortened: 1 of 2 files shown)


zbee (approved):
Fucking `?` operator

Tartarga (approved):
Yup, I didn't add a ! after converting...

Taurenkey:
> Yup, I didn't add a ! after converting...

Wasn't the issue, it's that if the status is null (ie, not on the target) it will evaluate as false because then it doesn't have an object to evaluate the remaining time variable on. That's why the previous check was both a 'if null' and a 'if less than time', both were required before the status changes.

edewen:
> > Yup, I didn't add a ! after converting...
> 
> Wasn't the issue, it's that if the status is null (ie, not on the target) it will evaluate as false because then it doesn't have an object to evaluate the remaining time variable on. That's why the previous check was both a 'if null' and a 'if less than time', both were required before the status changes.

Flan was referring to the sage part. It is not looking for a time. Simply for existing shield. But the way it is written, it was looking for  an existing shield in order to add another shield instead of looking for the absence of the shield.

Kagekazu (approved):

PR #498, opened 2025-04-16
[Core] Accept job parameters to the `list` commands

zbee:
- [X] Adds support for specifying a job, for any of the list commands - set, unset, or all


WrathCombo/Commands.cs
```diff
@@ -613,6 +663,7 @@ private void HandleOpenCommand
         }
 
         ConfigWindow.IsOpen = true;
+        ConfigWindow.OpenWindow = OpenWindow.PvE;
```
Tartarga, line 666: Random thought
if area is PvP OpenWindow.PvP else OpenWindow.PvE?

Tartarga (approved):

PR #499, opened 2025-04-16
[Core] New setting: always open Wrath to the PvE Features tab

### LINKED ISSUE #492: Open PVE Features UI to Current Job on Opening is broken
/wrath brings up the last selected menu item and is not navigating to current job in PVE as implied, this includes using the plugin installer menu

zbee:
- [X] Adds a new setting to allow `/wrath` command to always open to the PvE tab instead of the last tab.\
      (can still be overridden, like with `/wrath pvp` or `/wrath settings`)

Closes #492


Kagekazu (approved):

zbee:
@Tartarga I've just now added that as a separate option

Tartarga (approved):

PR #503, opened 2025-04-18
[SMN] Bozja Implementation

NebulousByte:
Basic draft of Bozja stuff for SMN. Completely untested, but based on existing code for GNB.


edewen:
As this has sat stagnant now for over 2 months, conflicts greatly with the current iteration of smn, im a close this for now. Not a terrible idea but there is better forms of special content coming eventually.

PR #516, opened 2025-05-01
[Target] Full Class Revision

Kaeris-Tempest:
### 1.1. New Function: IsTargetInRange()
- Uses squared distance instead of Sqrt() for range checks.
- Maintains the exact same precision as Sqrt().
- ✅ Up to **10x** faster.

### 1.2. New Function: GetTargetDistanceSquared()
- Provides hitbox-adjusted squared distance without Sqrt().
- Allows distance sorting for a fraction of the cost.

### 1.3. Refactor: All Sqrt-Based Distance Functions
- Migrated to squared distance where applicable (e.g., InMeleeRange, InActionRange).
- Simple distance checks: `GetTargetDistance() <= range` to `IsTargetInRange(range)`.
- Sorted distance checks: `GetTargetDistance()` to `GetTargetDistanceSquared()`.

### 1.4. Optimize: IsInLineOfSight()
- Uses accelerated NET 9 functions for normalization and magnitude.
- Removed ClientStructs.Vector3 ↔ Numerics.Vector3 conversions.
- ✅ Up to **2x** faster until raycast.

### 1.5. Optimize: GetTargetHPPercent() & PlayerHealthPercentageHp()
- Improved precision by multiplying before dividing (explicit cast no longer required).
- Skips calculations for dead / full HP (common cases).
- ✅ Up to **2–3x** faster for these states.

### 2.1. Refactor: AngleToTarget()
- Replaced branches with a `switch` structure and unified angle normalization.
- Now also powers OnTargetsRear(), OnTargetsFlank() and OnTargetsFront().
- ✅ Up to **20%** faster.

### 2.2. Optimize: CanCircleAoe(), CanRangedCircleAoe(), CanConeAoe(), CanLineAoe()
- Cached values where applicable to avoid repeated access when parsing.
- Simplified lambda logic (now branchless).
- ✅ Up to **40%** faster.

### 2.3. PositionalMath Library
- Compile-time constants for radian/degree conversions to avoid repeated divisions.
- Removed redundant float ↔ double ↔ float recasts across several functions.
- Now uses `MathF` to match `float` type used in other math operations.

### 3.1. Optimize: `Null` Checks
- Added null checks for certain functions that tried accessing parameters directly.
- Combined several null checks for faster early-exits in certain functions.

### 3.2. Added: XML `<summary>` Tags
- Every function in `Target.cs` now has a brief description of its functionality.
- Outdated or incorrect descriptions were also updated.

### 3.3. Optimize: AutoRotation
- The controller now benefits from the faster distance checks.
- Distance checks will now also occur ***before*** expensive line of sight checks.

### 3.4. Optimize: ActionWatching
- Simplified a few collections.
- Improved certain dictionary lookup functions.

### DIFF
--- a/WrathCombo/Combos/PvE/RDM/RDM_Helper.cs
+++ b/WrathCombo/Combos/PvE/RDM/RDM_Helper.cs
@@ -188,7 +188,7 @@ internal partial class RDM
                 || (GetRemainingCharges(Engagement) == 1 && GetCooldownRemainingTime(Engagement) < 3))
             && LevelChecked(Engagement)
             && !JustUsed(Engagement, 2f) //try not to double use, wait for next GCD
-            && IsTargetInRange(3f))
+            && InMeleeRange())
             newActionID = Engagement;
 
         if (newActionID == 0
@@ -257,7 +257,7 @@ internal partial class RDM
                 && !HasStatusEffect(Buffs.Dualcast)
                 && !HasStatusEffect(Role.Buffs.Swiftcast)
                 && !HasStatusEffect(Buffs.Acceleration)
-                && (IsTargetInRange(3f) || (GapCloser && HasCharges(Corpsacorps))))
+                && (InMeleeRange() || (GapCloser && HasCharges(Corpsacorps))))
             {
                 //Situation 1: Manafication first
                 if (DoubleCombo
@@ -373,7 +373,7 @@ internal partial class RDM
             bool MeleeEnforced = true)
         {
             //Normal Combo
-            if (IsTargetInRange(3f) || MeleeEnforced)
+            if (InMeleeRange() || MeleeEnforced)
             {
                 if ((ComboAction is Riposte or EnchantedRiposte)
                     && LevelChecked(Zwerchhau)
@@ -404,7 +404,7 @@ internal partial class RDM
             {
                 if (GapCloser
                     && ActionReady(Corpsacorps)
-                    && !IsTargetInRange(3f))
+                    && !InMeleeRange())
                 {
                     newActionID = Corpsacorps;
                     return true;
@@ -443,7 +443,7 @@ internal partial class RDM
                         return true;
                     }
                 }
-                if (IsTargetInRange(3f))
+                if (InMeleeRange())
                 {
                     newActionID = OriginalHook(Riposte);
                     return true;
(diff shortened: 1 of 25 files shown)


WrathCombo/Combos/PvE/RDM/RDM_Helper.cs
zbee, line 1: All these `3`s should be `InMeleeRange()`, that's what is actually being checked

WrathCombo/Combos/PvE/RDM/RDM_Helper.cs
```diff
@@ -456,15 +454,15 @@ internal static bool TrySTMeleeStart(ref uint newActionID,                 //Sim
         }
 
         internal static bool TryAoEManaEmbolden(ref uint newActionID,                 //Simple Mode Values
-            int MoulinetRange = 6)//idk just making this up
+            float MoulinetRange = 6f)//idk just making this up
```
zbee, line 457: It's actually 8y

WrathCombo/Combos/PvP/RPRPVP.cs
zbee, line 1: This is what I was talking about yesterday; doing `GetTargetDistance` once seems very likely to be more efficient than doing `IsTargetInRange` 5 times.
May well be worth keeping it how it was when it's done multiple times like this.

Kaeris-Tempest: You'd have to call `IsTargetInRange` about ten times to be around as costly as one `GetTargetDistance`. This is the worst case scenario and it's still twice as fast.


But this one in particular is a tricky use-case because the logic itself isn't structured very well, with proper flow you'd only need about 4 checks at most (Melee/5, Cone/8, Line/15, Cast/25), all of which can be cached.


The PvP role actions (like Smite here) should ideally have their own dedicated external functions with proper distance/status checks so they don't need to be done for every job (but that's a job for another time).

WrathCombo/Combos/PvE/RDM/RDM_Helper.cs
Tartarga: There is a slider attached for advanced on how close you want to be to do melee aoe. 6 I think was a older place holder because (I think) people wanted to be closer but not really close? Idk autism

Kaeris-Tempest: Set both 'moulinetRange' variables to 8f for the time being.

WrathCombo/Combos/PvE/RDM/RDM_Helper.cs
Kaeris-Tempest: Sorted. Also cached a lot of the `IsTargetInRange(1f)` calls in both `DRG.cs` and `WAR.cs` so it's a bit cleaner.

Kaeris-Tempest:
Going back in the oven for a bit.

PR #519, opened 2025-05-02
[BLM] Here we go again

Kagekazu:
- Cleanup Gauge stuff
- Fix `Swiftcast` - `Triplecast` doubleweaving
- Fix using `Fire4` instead of `Paradox`
- Add `Amplifier` - `Xenoglossy` combo
- Fix stuck on `Blizzard 4` when swapping from AoE to ST
- Make `Thunder` and `Thunder 2` not overwrite eachother


Akechi-kun (approved):
image

PR #521, opened 2025-05-03
[AST] Add option to use balance priorities for card quick target

Cthulhu252:
Add fixed priorities for cards and a combo box on the option to use them or keep the existing random selection
Priorities are based on the 2m windows

### DIFF
--- a/WrathCombo/WrathCombo.csproj
+++ b/WrathCombo/WrathCombo.csproj
@@ -56,7 +56,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <Content Include="..\res\plugin\wrathcombo.png" Link="images\wrathcombo.png" CopyToOutputDirectory="PreserveNewest" Visible="false"/>
+        <Content Include="..\res\plugin\wrathcombo.png" Link="images\wrathcombo.png" CopyToOutputDirectory="PreserveNewest" Visible="false" />
     </ItemGroup>
 
     <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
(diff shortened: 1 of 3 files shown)


WrathCombo/WrathCombo.csproj
zbee, line 1: Are the project file changes relevant to the code changes?

Cthulhu252: Thanks zbee, csproj should be showing unchanged again now

zbee (approved):
I would like to see a different priority list for Balance vs Spear, but this is definitely a lot better than randomness, so still a big win either way.

Cthulhu252:
> I would like to see a different priority list for Balance vs Spear, but this is definitely a lot better than randomness, so still a big win either way.

I was thinking of the best way to show this, because of the way the rest of the selection criteria works it prioritised ranged/melee for the relevent card. I could list SAM and PCT both as 1, but I thought that was more confusing than having a sequensed list, but both would give exactly the same result

zbee:
> both would give exactly the same result.

Perhaps I misunderstand the code, but I did mean essentially "prioritize melee for the balance" and "prioritize ranged for the spear", which I don't think it currently does.
Of course, correct me if I'm wrong, but like I said: it's still better than 100% rng that is currently there, so it's still fine by me even if you don't do this.

Cthulhu252:
> Perhaps I misunderstand the code, but I did mean essentially "prioritize melee for the balance" and "prioritize ranged for the spear", which I don't think it currently does.

It is doing that, lines 245 of the existing code (274 after my changes) checks to see if the selected player is suitable for the currently drawn card, if not it checks the next player. The way i have people sorted, if the card is spear then it will check the melees but they all fall into the unsuitable backup code, so the highest prio ranged will always get the spear if a ranged exists

zbee:
Nice - good stuff then

PR #522, opened 2025-05-03
[Core] Mouseover cleanup

Tartarga:
* Cleaned up and moved MouseOver Party UI code to PartyTargetingService, simplified functions. 
* Updated UITarget pointer to FFXIVClientStructs declared name instead of manually specifying
* Removed GetMouseOverHealTarget from Target, calls PartyTargetingService


Tartarga:
Part of New PR

PR #526, opened 2025-05-05
[SMN] Priority selection for Summon Order

edewen:
- [x] Added Priority Selection for Egi order following WHM example. 
- [x] Reordered ccp to actually make sense following down the list of the rotation


zbee:
Just FYI, I'll hop into this code to refine the UI and reduce option count if possible, and will do a review at that time.

Genesis-Nova (approved):
Looks good to me.

zbee (approved):

PR #534, opened 2025-05-10
[PVP] Fix Out-of-Range Purify issues

zbee:
- [X] Quick-Purify Statuses are now completely based off of a master list,\
      no possibility for out-of-range errors anymore.
  - [X] Quick-Purify Statuses UI and Logic are no longer hard-coded, instead just looping over that master list.

### DIFF
--- a/WrathCombo/Combos/PvP/ALL/ALL.cs
+++ b/WrathCombo/Combos/PvP/ALL/ALL.cs
@@ -1,6 +1,8 @@
-﻿using Dalamud.Game.ClientState.Objects.Types;
+﻿using System;
+using Dalamud.Game.ClientState.Objects.Types;
 using ECommons.GameHelpers;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


Kaeris-Tempest:
It shouldn't affect your changes, but `Sleep` and `Half-asleep` aren't present in PvP anymore; the ability that used to cause those was removed a few patches back.

zbee:
Well, then there should be a reset of the Quick Purify Statuses config, and the master list should have those removed, ... and we would need to add a way like Preset resetting to reset configs.
But, as you said, they can also just be left as they are ... and that sounds like a bother. I'll add a todo note.

Kaeris-Tempest:
Aye, I think it's fine to leave it as it is and there shouldn't be issues. It's not like the statuses were removed from the game files and the IDs are invalid. They still exist, there's just nothing that players can use to apply them currently.

PR #537, opened 2025-05-13
[Core] Add Hidden Features

zbee:
- [X] Adds Setting to show Hidden Features (on the Debug tab)
- [X] Adds Attribute to mark Presets as Hidden Features
- [X] Adds logic to only show Hidden Feature Presets when the Setting is enabled
- [X] Adds some DRK Hidden Features (for R6S and R7S)
- [X] Adds the same WAR Hidden Features

> This is not actually a draft, but I really don't want this to be merged until I step back into Savage with my group on Sunday, to make 100% certain this works exactly how it did before it was converted to Hidden Features.

<sub>Continues #531</sub>


zbee:
Tomorrow I will be able to test the Hidden Features I added, On Hold until that time.

zbee:
Tested the same as before in raid 👍🏻

zbee:
@Akechi-kun I would much appreciate it if you were able to pull and test this on WAR in non-Savage content, to double check that it doesn't fuck it up outside of Savage.
I've tested that it works in dungeons and Savage, and does the hidden features in Savage, but not necessarily checked fully that each feature with a hidden feature on it still functions correctly outside of the hidden feature's conditions.

That would include Interrupts, Stuns, and Reprisal+Shake, and I think that's it.

Kaeris-Tempest (approved):

PR #543, opened 2025-05-15
[Core] Status Cache Consolidation

Tartarga:
- Removed Sloth-era "Blocking Statuses". Should be covered by ActionReady
- Consolidated most Status functions under StatusCache.cs
- Redid most Invincibility lookups, 
- Added NumberOfEnemiesInRange function


WrathCombo/Data/StatusCache.cs
```diff
@@ -0,0 +1,245 @@
...
+        private static readonly HashSet<uint> AmnesiaStatusIds = InitializeStatusIdCaches()[GetStatusName(5)];
+        public static bool HasAmnesia() => HasStatusInCacheList(AmnesiaStatusIds);
+
+        /// <summary>
+        /// A cached set of status IDs for the "Pacification" status effect (name ID 6).
+        /// </summary>
+        /// <remarks>
+        /// Populated at startup. Status IDs are stable across languages.
+        /// </remarks>
+        private static readonly HashSet<uint> PacificationStatusIds = InitializeStatusIdCaches()[GetStatusName(6)];
```
zbee, line 103: `GetStatusName` is already done at line 60-70. Maybe those should be taken out of there and here, and put in a separate variable to be done just once?
I assume both of these places it will only ever do it once, but am not sure, and I know that lookups like that are not necessarily speedy, so it's just a thought.

zbee: I also realize now that it's running a new `InitializeStatusIdCaches()` every time ... that should definitely be run once, and left in a variable.

WrathCombo/CustomCombo/Functions/Status.cs
```diff
@@ -203,74 +170,53 @@ public static bool NoBlockingStatuses(uint actionId)
...
+        //        (alliance != AllianceGroup.GroupC && targetStatuses.Contains(2411)))
+        //        return true;
+
+        //    // Omega
+        //    if ((targetStatuses.Contains(1674) || targetStatuses.Contains(3454)) && (HasStatusEffect(1660) || HasStatusEffect(3499)) ||
+        //        (targetStatuses.Contains(1675) && (HasStatusEffect(1661) || HasStatusEffect(3500))))
+        //        return true;
+
+        //    return false;
+        //}
```
zbee, line 220: I think best to just remove this entirely - it will be in git history and `StatusCache`, no need.

WrathCombo/CustomCombo/Functions/Status.cs
```diff
@@ -97,98 +95,67 @@ public unsafe static float GetStatusEffectRemainingTime(ushort effectId, IGameOb
...
+        //        }
+        //    }
+        //    catch (Exception ex)
+        //    {
+        //        Svc.Log.Error(ex, $"Error checking for cleansable debuff on target {(target?.Name ?? "null")}");
+        //        return false;
+        //    }
+
+        //    return false;
+        //}
```
zbee, line 158: Again, I think best just remove this altogether.

WrathCombo/CustomCombo/Functions/Status.cs
Tartarga: It'll happen when i get it in a good place for final review, just easier to go back and quickly look

WrathCombo/CustomCombo/Functions/Status.cs
```diff
@@ -203,74 +170,53 @@ public static bool NoBlockingStatuses(uint actionId)
                 case ActionWatching.ActionAttackType.Ability:
                     if (HasAmnesia()) return false;
                     return true;
-
+                case ActionWatching.ActionAttackType.Unknown:
+                    break;
```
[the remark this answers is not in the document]
Tartarga: I was trying to shutup suggestions XD

WrathCombo/Data/StatusCache.cs
```diff
@@ -52,92 +52,65 @@ internal class StatusCache
...
-                    Svc.Log.Warning($"No status IDs found for status name: {name}");
-            }
-
-            return caches;
+            Amnesia = 5,
+            Pacification = 6,
+            Silence = 7,
+            Weakness = 43,
+            BrinkOfDeath = 44,
+            DamageDown = 62
```
zbee, line 64: Does this work on non-english?

Tartarga: Yes

Tartarga:
Okay I'm at a good spot for now I think. hit me up with more questions

WrathCombo/Data/StatusCache.cs
zbee: I'm surprised

Tartarga:
```
var statusEffects = Enum.GetValues<StatusEffect>();
var nameToEffect = statusEffects
    .Select(e => (Effect: e, Name: GetStatusName((uint)e)))
```
ex Amnesia's value of 5 is looked up to a name (string) like the old method, then pulls all "Amnesia" strings from sheets

Might go icon route testing though...

Kagekazu (approved):

zbee (approved):
since it's been working for Kage for near a week, no issues

Kagekazu (approved):
Been using for a week without issues, good for me!

PR #544, opened 2025-05-16
[DRG] Lifesurge fix

Kagekazu:
- fix lifesurge for all lvls

### DIFF
--- a/WrathCombo/Combos/PvE/DRG/DRG_Helper.cs
+++ b/WrathCombo/Combos/PvE/DRG/DRG_Helper.cs
@@ -24,6 +24,12 @@ internal partial class DRG
                  JustUsed(FangAndClaw) ||
                  JustUsed(OriginalHook(VorpalThrust)) && LevelChecked(HeavensThrust)))
                 return true;
+
+            if (!LevelChecked(Drakesbane) && JustUsed(VorpalThrust))
+                return true;
+
+            if (!LevelChecked(FullThrust) && JustUsed(TrueThrust))
+                return true;
         }
 
         return false;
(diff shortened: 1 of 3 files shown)


WrathCombo/Combos/PvE/DRG/DRG_Helper.cs
```diff
@@ -24,6 +24,12 @@ internal static bool UseLifeSurge()
                  JustUsed(FangAndClaw) ||
                  JustUsed(OriginalHook(VorpalThrust)) && LevelChecked(HeavensThrust)))
                 return true;
+
+            if (!LevelChecked(Drakesbane) && JustUsed(VorpalThrust))
+                return true;
+
+            if (!LevelChecked(FullThrust) && JustUsed(TrueThrust))
+                return true;
```
Kaeris-Tempest, line 32: This works but only down to level 18; because `UseLifeSurge()` is only called inside `HasStatusEffect(Buffs.PowerSurge)`. Disembowel is the earliest way to get `Power Surge` at level 18.

You can make it work all the way down to level 6 (as soon as you learn `Life Surge`) by changing `HasStatusEffect(Buffs.PowerSurge)` into `HasStatusEffect(Buffs.PowerSurge) || !LevelChecked(Disembowel)`.

So then it's either:
- You have the damage buff (implicit level check) **OR**
- You can't have it yet (not learned).

Kagekazu: Oh yeah, forgot about disembowel lvl.
Added!

zbee (approved):
Good switches to `InMeleeRange()`

PR #549, opened 2025-05-21
[RPR] Some cleanup

Kagekazu:
- some SoD Check update


Kagekazu:
Doublechecked, is unneeded

PR #550, opened 2025-05-22
[BRD] Update to nowaste Handler

edewen:
- [x] Added all content vs Non Bosses Radial to ST buffs, ST dots, and AoE buffs
- [x] Added health threshold to the same. Up to 100% for trash disabling
- [x] Removed health threshold from songs as they were written to be target neutral months ago
- [x] Removed no waste option from CCP as it is now baked into the buffs and dots options
image

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -771,6 +771,10 @@ public enum CustomComboPreset
     BRD_ST_Adv_Balance_Standard = 3048,
 
     [ParentCombo(BRD_ST_AdvMode)]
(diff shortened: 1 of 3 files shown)


zbee:
I think the UI option of "Non-Bosses vs All Enemies" makes more sense here than "All Enemies vs Boss only", as I don't think there is a case where you would want this on Bosses but NOT trash mobs; which is what Kage showed:
image

https://github.com/PunishXIV/WrathCombo/blob/57b1757e757f5da342e22ea4a93b172c2919b7c7/WrathCombo/Combos/PvE/BLM/BLM_Config.cs#L62-L69

edewen:
Updated to the blm PR style as requested.

zbee (approved):

PR #556, opened 2025-05-25
[PCT] 7.2 Pictoverhaul

edewen:
What it needs at this point. Is just lots of live testing and logs. So I can minimize the drift and make sure it is consistent. But I don't raid with it or main it. 

- [x] Advanced st
- [x] Advanced aoe
- [x] Simple st
- [x] Simple aoe
- [x] CCP/UI Cleanup
- [x] All Regions set up

Updated Single Target Advanced to use the h2/h3 burst as well as clean some things up with adding some small pooling to make h2/3 work. As such, removed the choose your burst window spells options. Simply a choice on whether to use the optimized burst window. It won't function if you disabled half the spells.

Added Definitions to helperfile as well as burst window function and Hyperfantasia function.

Updated Aoe to match the single target setup retaining the holy in white option as it is a dps increase to not hold them
There is no optimized burst window option for aoe. 

Converted the advanced modes to simple modes. 

Cleaned up and reorganized the ccp and ui. Removed un-used options and reorganized in a logical manner. 

Hyperfantasia  movement "option" was not actually hooked up from the checkbox to the code. It doesnt really need to be. it is just using holy or comet before hammer in this very specific event. removed the unused checkboxes. The hyperfantasia priority still functions. Just baked into ccp description.

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -3,6 +3,7 @@ using System;
 using WrathCombo.Attributes;
 using WrathCombo.Combos.PvE;
 using WrathCombo.Combos.PvP;
+using static WrathCombo.Combos.PvE.PCT;
 namespace WrathCombo.Combos;
 
 /// <summary> Combo presets. </summary>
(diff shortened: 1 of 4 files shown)


zbee (changes requested):
With this level of change to the `CustomComboPreset` for a job, I would want to see a Reset of user options, or using the old IDs where possible and adding new IDs for new options instead of re-using IDs for different features.
When re-use of IDs is more minor, I can't say I care much; but this seems to re-use quite a few for new Presets, _as well as_ keeping some Presets but giving them an ID of a now-removed and unrelated Preset.

With it being a 7.2 update, and something of a rework, I think either of those is acceptable, so I'll leave it up to you which to do.
Sorry I didn't notice it sooner.
Of course reach out if you're not sure how to do a Reset, if that's what you want to do.

PR #557, opened 2025-05-25
[Data] ReceiveActionEffectDetour Null Fix

Kaeris-Tempest:
- Fixes the error caused by using an action on an object not in the player's object list.
- Refactored `TimeSinceActionUsed` and `TimeSinceLastSuccessfulCast` with `TryGetValue`.


zbee (approved):
As it was already thoroughly tested

PR #565, opened 2025-05-26
[Target] Minor Performance Improvements

Kaeris-Tempest:
- Added references to `GetTargetDistance` and `GetTargetHeightDifference` so `LocalPlayer` is only fetched once.
- Rewired `InMeleeRange()` to avoid unnecessary `double` promotion; `Service.Configuration.MeleeOffset` is stored as `double` but only ever retains a `float` value, since input comes from `ImGui.InputFloat`.
- Optimized `GetTargetHPPercent` to perform division last; no longer requires explicit `(float)` cast and now matches the accuracy of `PlayerHealthPercentageHp()`.


zbee:
Thank you 🙏

PR #566, opened 2025-05-26
[IPC] Handle IPv6 better

zbee:
- [X] Uses Dalamud's `Happy Eyeballs` handler for the HTTP client used to check the status of the IPC
- [X] Adds a short timeout, if that still doesn't handle some peoples' network setups
- [X] Applies the same to the MotD http client


zbee:
Would appreciate if anyone could pull and build this, and just make sure the MotD appears upon re-log.
If it doesn't: I'll need those errors.

PR #585, opened 2025-05-30
[PLD] Fix Fight or Flight Blocking AoE Modes

Kaeris-Tempest:
Issue was somewhat obscure, but should now be resolved for good.

Culprit(s):
- `ActionReady` has an embedded `OriginalHook`, checking whatever the action changed into.
- The game's native action change function affects this.
- Directly calling an action that has changed into something else will hang the logic.

```
// Blocked by Goring Blade
// Fight or Flight is not ready, but Goring Blade is
if (ActionReady(FightOrFlight))
	return FightOrFlight;
```
```
// Resolved
// We ensure Fight or Flight hasn't changed into anything else first
if (OriginalHook(FightOrFlight) is FightOrFlight && ActionReady(FightOrFlight))
	return OriginalHook(FightOrFlight);
```

### DIFF
--- a/WrathCombo/Combos/PvE/PLD/PLD.cs
+++ b/WrathCombo/Combos/PvE/PLD/PLD.cs
@@ -39,6 +39,7 @@ internal partial class PLD : Tank
 
             #region Variables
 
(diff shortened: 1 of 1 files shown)


zbee (approved):
`OriginalHook` be like

PR #594, opened 2025-05-31
[Retargeting] Fix `AnyDeadNonPartyMember`

zbee:
- [X] Fixes the `IsInParty` `IGameObject` extension
- [X] Fixes the `AnyDeadNonPartyMember` `SimpleTarget` (non party members are not necessarily `IBattleChara`)
- [X] Fixes `IsDeadEnoughToRaise` (same as above)


Taurenkey:
> Fixes the AnyDeadNonPartyMember SimpleTarget (non party members are not necessarily IBattleChara)

What do you mean by that? By definition all player characters are also battle characters.

edewen (approved):
Tested it on me anyway

PR #600, opened 2025-06-01
[BRD] Added missing retargeted in ccp

edewen:
Added missing retargeted  flag in ccp for Wardens party cleanse


zbee (approved):
Thanks for doing that

PR #602, opened 2025-06-01
[PLD] Shield bash spam fix

edewen:
- [x] Removed the forced shield bash from simple. 
- [x] Turned shield bash into its own option for advanced.
- [x] Tested and went down a rabbit hole caused entirely by Zbee and no one else. 

Just a quick fix cause I was bored. Obviously if you wanna do your own shit @Kaeris-Tempest I can close this


zbee (approved):
Approved. Did also add a `JustUsedOn` to `ShieldBash` usage, to actually address the spam, instead of only putting it behind an option.

Would like to see it remain in the Simple Modes with that change, but doesn't matter either way to me.

PR #603, opened 2025-06-01
[PCT] Tweaks and Hammer fix

edewen:
Hammer was getting stuck in the burst window when someone would skip starry muse locking it up on hammer stamp 1 even though it was already cast. 

Added a buff check on top of the existing justused ( still needed for propagation purposes) to push it into next gcd

added a || !canweave to starry muse so that if the delayed weave is missed due to the slighest lag, it still pushes the 2 min instead of adding s more seconds of drift. (lesser of 2 evils)

Added charge check on steelmuse because if run without the opener, the first windows starry is too close to steel in cooldown, leading to it hanging on steel muse for couple second at 2 min window before fixing itself for the 4 min.

### DIFF
--- a/WrathCombo/Combos/PvE/PCT/PCT.cs
+++ b/WrathCombo/Combos/PvE/PCT/PCT.cs
@@ -40,7 +40,7 @@ internal partial class PCT : Caster
             #endregion
 
             #region Burst Window lvl 100 only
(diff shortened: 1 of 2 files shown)


Kagekazu (approved):
Logic seems sound, havent tested in game tho.

PR #604, opened 2025-06-02
[WHM] Status Effect Cap Code +Pom Visual fix

edewen:
When popping Presence of mind, The Actionready picks up the original hook ready from glare 4, and when weaving, it shows a greyed out Pom. 

This fixes that weird looking attempt at weave during the Pom Burst window. 

Added New status cap code to needsdot()


zbee (approved):

Kagekazu (approved):
Tested in game, no issues.

PR #605, opened 2025-06-02
[BRD] Status Cap Code

edewen:
Added Status cap code

Kind of hard to actively test so did some creative additions of ! to ensure the Logic in bard code is at least functioning


Kagekazu (approved):
LGTM

PR #606, opened 2025-06-02
[BLM] Polyglot usage tweaks

nativepapaya:
Initial work on making polyglot usage a bit more flexible.

- Added a slider that allows user to select how many chargers to keep for manual use, seperate from the movement option slider
- Removed the "USE ASAP" option as slider can be set to "0" to recreate the same behavior
- Make the movement slider and manual usage slider not conflict with each other.
- Added priority movement options

### DIFF
--- a/WrathCombo/Combos/PvE/BLM/BLM.cs
+++ b/WrathCombo/Combos/PvE/BLM/BLM.cs
@@ -258,13 +258,8 @@ internal partial class BLM : Caster
             if (IsEnabled(CustomComboPreset.BLM_ST_UsePolyglot))
             {
                 //Overcap protection
-                if (HasMaxPolyglotStacks && PolyglotTimer <= 5000)
-                    return LevelChecked(Xenoglossy)
-                        ? Xenoglossy
-                        : Foul;
-
-                if (IsEnabled(CustomComboPreset.BLM_ST_UsePolyglotAsap) &&
-                    HasPolyglotStacks())
+                if (HasMaxPolyglotStacks && PolyglotTimer <= 5000 || 
+                    PolyglotStacks > BLM_ST_Polyglot_Save)
                     return LevelChecked(Xenoglossy)
                         ? Xenoglossy
                         : Foul;
@@ -324,8 +319,7 @@ internal partial class BLM : Caster
             {
                 // TODO: Revisit when Raid Buff checks are in place
                 if (IsEnabled(CustomComboPreset.BLM_ST_UsePolyglot) &&
-                    ((BLM_ST_MovementOption[3] && PolyglotStacks > BLM_ST_Polyglot_Movement) ||
-                     (!BLM_ST_MovementOption[3] && HasPolyglotStacks())))
+                    BLM_ST_MovementOption[3] && PolyglotStacks > BLM_ST_Polyglot_Movement)
                     return LevelChecked(Xenoglossy)
                         ? Xenoglossy
                         : Foul;
--- a/WrathCombo/Combos/PvE/BLM/BLM.cs
+++ b/WrathCombo/Combos/PvE/BLM/BLM.cs
@@ -257,7 +257,7 @@ internal partial class BLM : Caster
                 return Scathe;
 
             //Overcap protection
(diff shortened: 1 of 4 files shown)


zbee:
From your TODO this sounds like it maybe should still be a draft. If so, you can convert it near the top right.

Kagekazu (approved):
All works as it should.

WrathCombo/Combos/PvE/BLM/BLM_Config.cs
```diff
@@ -53,17 +57,26 @@ internal static void Draw(CustomComboPreset preset)
...
 
                 case CustomComboPreset.BLM_ST_UsePolyglot:
-                    if (BLM_ST_MovementOption[3])
-                        DrawSliderInt(1, 3, BLM_ST_Polyglot_Movement,
-                            "How many charges to save for movement?");
+                    DrawSliderInt(0, 3, BLM_ST_Polyglot_Save,
+                        "How many charges to save for manual use?");
+
+                    DrawSliderInt(0, 3 - BLM_ST_Polyglot_Save, BLM_ST_Polyglot_Movement,
+                        "How many charges to save for movement?");
```
zbee, line 75: This will allow users to set it to save 3 for movement, and then also 3 for manual use.
At the very least, `BLM_ST_Polyglot_Save`'s slider should also subtract `BLM_ST_Polyglot_Movement`, but -as discussed- an `if` that sets the other value lower if need-be is safer.

WrathCombo/Combos/PvE/BLM/BLM_Helper.cs
```diff
@@ -53,6 +53,50 @@ internal static readonly Dictionary<uint, ushort>
 
     internal static bool HasPolyglotStacks() => PolyglotStacks > 0;
 
+    #region Movement Prio
+
+    private static (uint Action, CustomComboPreset Preset, System.Func<bool> Logic)[]
+        PrioritizedMovement =>
```
zbee, line 59: Oh hey I recognize this priority methodology 😆

WrathCombo/Combos/PvE/BLM/BLM.cs
```diff
@@ -292,40 +285,22 @@ protected override uint Invoke(uint actionID)
...
             if (FirePhase)
             {
                 // TODO: Revisit when Raid Buff checks are in place
                 if (IsEnabled(CustomComboPreset.BLM_ST_UsePolyglot) &&
-                    ((BLM_ST_MovementOption[3] && PolyglotStacks > BLM_ST_Polyglot_Movement) ||
-                     (!BLM_ST_MovementOption[3] && HasPolyglotStacks())))
+                    BLM_ST_MovementOption[3] &&
+                    PolyglotStacks > BLM_ST_Polyglot_Movement ||
+                    !BLM_ST_MovementOption[3] &&
+                    PolyglotStacks > BLM_ST_Polyglot_Save)
```
zbee, line 303: Very possibly incorrect, but I think this needs a set of `()` around the first two items, and then around the second two items, and possibly then around both of those sets, for this `||` to be correct, but I didn't test in-game.
As in:
```cs
((BLM_ST_MovementOption[3] &&
  PolyglotStacks > BLM_ST_Polyglot_Movement) ||
 (!BLM_ST_MovementOption[3] &&
  PolyglotStacks > BLM_ST_Polyglot_Save)))
```

zbee (changes requested):
Looks like some minor changes may be desirable, but otherwise the logic and changes look good; didn't test in game.

WrathCombo/Combos/PvE/BLM/BLM.cs
Kagekazu: That is correct, i pushed a fix for it!

WrathCombo/Combos/PvE/BLM/BLM_Config.cs
zbee: Ah I see that ...
So, that would work, but is going to be checked constantly with the combo, when it only needs checked at the time that one of those configs is changed.

If you instead do that same code in an `if` on the slider config itself, it will only run when the config is changed.

PR #607, opened 2025-06-02
[UI] ActionReady and ActionSheet Debug Updates

Kaeris-Tempest:
- Added all `Phantom Actions`.
- Added a few missing PvP actions.
- Refactored `ActionSheet` to display actions based on what category they belong to.

### DIFF
--- a/WrathCombo/Window/Tabs/Debug.cs
+++ b/WrathCombo/Window/Tabs/Debug.cs
@@ -498,8 +498,9 @@ internal class Debug : ConfigWindow, IDisposable
         var actionsPvP = actionSheet
             .Where(x =>
                 x.IsPvP &&
-                x.Icon is not 405 &&
-                (x.ClassJobCategory.Value.IsJobInCategory(Player.Job) || x.ActionCategory.RowId is 0))
+                x.Icon != 405 &&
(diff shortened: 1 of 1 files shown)


Kaeris-Tempest:
Agreed on the renaming; I'll be making it `PvE`, `PvP` and `Occult` for now.
Could possibly also add `Bozja` and `Eureka` though we're a few years late for those by now...

Not sure what you meant on number 1? You mean set it to the `PvE` type by default, instead of having to pick a type?

Kaeris-Tempest:
- Added all `Bozja` actions.
- Renamed categories into `PvE`, `PvP`, `Bozja` and `Occult`.

I'm running into some weirdness when setting the default, I'll play safe and leave it `null` at the start.

Kaeris-Tempest:
- Added new `Cast Data` tree node under `Target Data` that shows extensive cast information.

Kagekazu (approved):
LGTM

Kaeris-Tempest:
Last few adjustments.
- Use `GetRowOrDefault` instead of `GetRow` to avoid any unwanted surprises.
- Now also displays `IsCastInterruptible` switch, because why not.

zbee (approved):
Still works as expected, love the changes, thanks

PR #608, opened 2025-06-02
[Retargeting] Fix DNC+AST+WHM

zbee:
- [X] Fixes DNC's bad Partner logic
- [X] Early-Returns DNC's Partner if only one valid one was found
      (from #621)
- [X] Separates DNC's level-based partner priority sorting, to prevent any conflicts between the sorts
      (from #621)
- [X] Removes AST's Card Resolver throttle
- [X] Clears Retargets when you die
- [X] Fixes to Retargeting to use the Retarget for ground-targeted abilities
- [X] Added options for whom to target with WHM's Asylum

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -7023,6 +7023,16 @@ SMN.JobID)]
     [Retargeted]
     WHM_AoEHeals_Asylum = 19028,
 
+    [ParentCombo(WHM_AoEHeals_Asylum)]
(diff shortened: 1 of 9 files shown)


Kaeris-Tempest (approved):

yurikoma (approved):
I have tested The WHM and AST changes. Both appear to be working nicely without issue.

WHM is correctly casting Asylum now based on setting chosen. I have witnessed it working correctly over 5 times with each setting, when set to target myself, target enemy or targeting focused ally target. Performed these tests in dungeons and 8-man raids

AST is no longer producing Card Target related errors in the dalamud logs. tested this over multiple instances and waiting for all party members to leave, also in some CE's in the new instanced area.

PR #610, opened 2025-06-02
[Core] Fix `StatusList` crashes

zbee:
- [X] Reset Auto Rotation's `OptionalTarget` immediately after use, not letting it get stale in the first place.
- [X] Remove accesses of `StatusList` at the same time an object is determined not to be an `IBattleChara`.
- [X] Check for `0` Statuses, which is how they are marked as `<nothing>`, not with a `null`.\
  (they're not even null-able)

### DIFF
--- a/WrathCombo/AutoRotation/AutoRotationController.cs
+++ b/WrathCombo/AutoRotation/AutoRotationController.cs
@@ -754,8 +754,21 @@ namespace WrathCombo.AutoRotation
 
             internal static bool CanAoEHeal(uint outAct = 0)
             {
-                var members = GetPartyMembers().Where(x => !x.BattleChara.IsDead && x.BattleChara.IsTargetable && (outAct == 0 ? GetTargetDistance(x.BattleChara) <= 15 : InActionRange(outAct, x.BattleChara)) && ((float)x.CurrentHP / x.BattleChara.MaxHp * 100) <= cfg.HealerSettings.AoETargetHPP);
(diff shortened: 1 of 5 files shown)


Kaeris-Tempest (approved):
`0` be like

yurikoma:
As a WHM main, I have been crashing somewhat regularly the past 3 days while while doing dungeons, or just fates and CE in South Horn, sometimes before a fate has even started, seems somewhat random but the event viewer has nearly always pointed to this issue being the problem.

have built and added this Fix, and have not yet received any more crashes since, could just be my luck suddenly changed, but due to the frequency of the crashes previously I am just going to say that this patch did its job and resolved the problem.

Will keep using the test build while waiting for this to officially get merged, and update here if anything changes on my end.

zbee:
@yurikoma Thanks so much for testing this, and for the error prior to the crash, that was very helpful.
I have pushed new changes to address that error, hopefully stopping the crash completely now.

zbee:
The try/catching before the most recent 3 commits did nothing, and has been removed, and the attempts to check the live-ness of the `IGameObject` have been replaced with the root fix of resetting `OptionalTarget`.
This change was extensively tested by @yurikoma in the [Discord Issue Post](https://discord.com/channels/1001823907193552978/1378784810625007668/1380901934600425533) for this problem, big thanks for that.

yurikoma (approved):
Have been testing these changes extensivly, I would previously crash after 2 in 3 instances however with this update I have completed over 50 instances without a single problem.

PR #613, opened 2025-06-02
[WAR] Add `Holmgang` Retargeting

zbee:
- [X] Added Feature to always Retarget `Holmgang` to self (as in, never let it go onto a target, which is pointless and can be dangerous in add-pulls)


Akechi-kun (approved):

edewen (approved):
Built. Tested. Mother Approved.

PR #615, opened 2025-06-03
[Melees] TN Cleanup + DRG Logic

Kagekazu:
DRG:
- Add Life Of the Dragon Checks to most OGCD's (so it uses burst with most buffs)
- Add failsafe so it force uses mirage dive when about to expire
- Add better logic for double mirage dive option

RPR:
- Add better TN check for Gluttony usage

All Melees:
- Remove obselete `CanDelayedWeave` checks on TN

### DIFF
--- a/WrathCombo/Combos/PvE/NIN/NIN.cs
+++ b/WrathCombo/Combos/PvE/NIN/NIN.cs
@@ -561,8 +561,8 @@ internal partial class NIN : Melee
             int bloodbathThreshold = 50;
             double playerHP = PlayerHealthPercentageHp();
             bool phantomUptime = true;
-            bool trueNorthArmor = TargetNeedsPositionals() && !OnTargetsFlank() && Role.CanTrueNorth();
-            bool trueNorthEdge = TargetNeedsPositionals() && !OnTargetsRear() && Role.CanTrueNorth();
+            bool trueNorthArmor = !OnTargetsFlank() && Role.CanTrueNorth();
+            bool trueNorthEdge = !OnTargetsRear() && Role.CanTrueNorth();
             bool dynamic = true;
 
             if (ActionWatching.TimeSinceLastAction.TotalSeconds >= 5 && !InCombat())


WrathCombo/Combos/PvE/DRG/DRG.cs
```diff
@@ -253,8 +258,10 @@ protected override uint Invoke(uint actionID)
                         ActionReady(MirageDive) &&
                         CanDRGWeave(MirageDive) &&
                         HasStatusEffect(Buffs.DiveReady) &&
-                        (IsEnabled(CustomComboPreset.DRG_ST_DoubleMirage) && LoTDActive ||
-                         IsNotEnabled(CustomComboPreset.DRG_ST_DoubleMirage)))
+                        (IsEnabled(CustomComboPreset.DRG_ST_DoubleMirage) &&
+                         TraitLevelChecked(Traits.LifeOfTheDragon) && LoTDActive ||
+                         IsNotEnabled(CustomComboPreset.DRG_ST_DoubleMirage) ||
+                         GetStatusEffectRemainingTime(Buffs.DiveReady) <= 3))
                         return MirageDive;
```
Akechi-kun, line 265: hmm, I would add rearrange it.
```
//Mirage Feature
if (IsEnabled(CustomComboPreset.DRG_ST_Mirage) &&
ActionReady(MirageDive) &&
CanDRGWeave(MirageDive) &&
HasStatusEffect(Buffs.DiveReady) &&
((IsEnabled(CustomComboPreset.DRG_ST_DoubleMirage) && 
(LoTDActive || GetStatusEffectRemainingTime(Buffs.DiveReady) <= 3f)) || //here instead because if enabled, this option is the only reason why the buff is being held
IsNotEnabled(CustomComboPreset.DRG_ST_DoubleMirage))) //if disabled, we just send immediately
    return MirageDive;
```

Probably makes virtually no difference, but the issue was more with the `Double Mirage Dive` option itself not having the failsafe, since with it disabled it will just send immediately after `Jump` and pretty much never fail.

WrathCombo/Combos/PvE/DRG/DRG.cs
```diff
@@ -311,14 +319,16 @@ FirstmindsFocus is 2 &&
...
                         return Starcross;
 
                     //Rise of the Dragon Feature
                     if (IsEnabled(CustomComboPreset.DRG_ST_Dives_RiseOfTheDragon) &&
                         ActionReady(RiseOfTheDragon) &&
                         CanDRGWeave(RiseOfTheDragon) &&
-                        HasStatusEffect(Buffs.DragonsFlight))
+                        HasStatusEffect(Buffs.DragonsFlight) &&
+                        LoTDActive)
                         return RiseOfTheDragon;
```
Akechi-kun, line 332: I personally don't think these need the LOTD checks, since `Dragonfire Dive` will now only be used during `LOTD` ensuring this will follow suit too since it's sometimes called right after, and `Starcross` is pretty much the same case with `Stardiver`.

Akechi-kun (approved):

WrathCombo/Combos/PvE/DRG/DRG.cs
Kagekazu: idk how long the buffs last, thats why i added it as a last resort, i can remove em after i check the buff length in game :D

WrathCombo/Combos/PvE/DRG/DRG.cs
Kagekazu: sure, makes sense! ill rearrange it

Akechi-kun (approved):

WrathCombo/Combos/PvE/ALL/Roles/RoleActions.cs
```diff
@@ -156,7 +156,7 @@ public static bool CanFeint() =>
             ActionReady(Feint) && !HasStatusEffect(Debuffs.Feint, CurrentTarget, true);
 
         public static bool CanTrueNorth() =>
-            ActionReady(TrueNorth) && TargetNeedsPositionals() && !HasStatusEffect(Buffs.TrueNorth);
+            ActionReady(TrueNorth) && TargetNeedsPositionals() && !HasStatusEffect(Buffs.TrueNorth) && CanDelayedWeave();
```
zbee, line 159: As you noticed in RPR, this does lead to some duplicate Weave checks, including in NIN SAM and VPR. Pretty irrelevant, but may want to delete those in the same PR as this.

zbee (approved):
Code changes make sense to me, and the TN changes all look correct; didn't test in game.

Tartarga:
Not sure if you know, but a few role actions like can head graze have a "what kind of weave to check" parameter if the melees need different weaves

zbee:
> if the melees need different weaves
True North will always be a Delayed Weave, to give the user a chance to move into position, that was standardized a while ago, and thankfully can now be forced.

PR #618, opened 2025-06-04
[SMN] Cleanup 2.0

edewen:
More adjustments for Summoner in line with Dawntrail Changes.

1. Removed the Burst Delay Option. This was needed when the 2 minute buff was 30 seconds long and popped on pull before the Demi. Now that it is 20 seconds and popped during demi. Setting it to 3 GCDS can very easily push OGCDS out of the demi summon window when pooling. Resulting in a need for the Emergency Demi Dump. Leaving that there for now.

2. Removed the any searing burst option checkbox. The code still looks for any searing, Just removed it as a choice. No other classes are concerned with 2 of the same class in a raid and set up several lines of code around it.Because it accounts for any burst. If another summoner IS with you, and you have demi only burst selected it will on its own start running with the next demi phase. 

3. Because it is Automatic, removed the selector for choosing which demi phase to burst with. This led to issues searing could get misaligned due to death or dungeon and be held off cooldown for over a minute. 

4. Added retargetting option to rekindle in advanced modes. 

5. Tired of users asking why summoner doesn't work just to find out it is the wrong ruin. No longer. It now applies to all 3 ruins. But wait there's more, For one small click they can use a radio button to choose the option from ruin1/2 only like before, set up just like astro. 

6. Added Drift adjustment for Searing light at slightly higher spellspeeds. Due to autorotation not using the queue, extra spellspeed/performance mode/ lower throttle is needed to ensure 6/6 Demi attacks in the window. This added spellspeed leads to lower cd on Demi which misaligns the Searing cd from the slightly short cd Demi. WIll automatically Fire an extra Ruin 3 to realign if the misalignment falls between 3 and 8 seconds. 

https://xivanalysis.com/fflogs/8XAhtWPT2Z3rpGwj/4/3

Overall adjustments to simplify the user end of Summoner as well as the code by removing unnecessary things that were based off 30 second buff in endwalker. 

- [x] Advanced ST
- [x] Advance AoE
- [x] Menu Cleanup
- [x] Ruin Choice
- [x] Playtesting and Analyzing
- [x] Covert Simple modes

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -5750,9 +5750,10 @@ public enum CustomComboPreset
     [CustomComboInfo("Rekindle Combo Option", "Adds Rekindle to the single target combo.", SMN.JobID)]
     SMN_ST_Advanced_Combo_DemiSummons_Rekindle = 17028,
 
-    [ParentCombo(SMN_ST_Advanced_Combo_DemiSummons_Attacks)]
-    [CustomComboInfo("Lux Solaris Combo Option", "Adds Lux Solaris to the single target combo.", SMN.JobID)]
-    SMN_ST_Advanced_Combo_DemiSummons_LuxSolaris = 17029,
+    [ParentCombo(SMN_ST_Advanced_Combo_DemiSummons_Rekindle)]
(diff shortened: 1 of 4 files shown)


zbee:
Is this still WIP? I assume not from being taken out of draft, but am not certain.

zbee (approved):
Tested in game and it performed at least as well as it used to for my level 90.
The UI for the class is much improved.

PR #620, opened 2025-06-05
[WAR] fix `Primal Wrath` triple-weaving

Akechi-kun:


edewen (approved):
Its a weave, didn't have a weave tag. Simple case. 

Though if canweave doesn't cut it, may need the !ActionWatching.HasDoubleWeaved() I aquaire many moons ago from MCH code.

Kagekazu (approved):
Gotta love dem weaves

PR #621, opened 2025-06-06
[DNC] Improve partner-priority-by-level sorting

### LINKED ISSUE #619: Desirable Dance Partner Feature is Using Lv90 Prio at Lv100
I'm noticing that when using the auto-Dance Partner feature it was picking targets that weren't in the correct priority to the Lv100 Balance choices. When looking through the code I saw the two functions:

`internal static readonly Dictionary<uint, int> Job100Prio = new()`
and
`internal static readonly Dictionary<uint, int> Job090Prio = new()`

And looking through both the dance partner seemed to be taking preference from the Job090Prio based on my testing so far (picking DRG over VPR in M7S one such case).

In the `bool TryGetBestPartner(out IGameObject? newBestPartner, int step = 0)` within DNC_Helper.cs I noticed the following snippet of code
```
filter = filter
                .OrderBy(x =>
                    PartnerPriority.RolePrio.GetValueOrDefault(
                        x.ClassJob.RowId.Role(), int.MaxValue))
                .ThenBy(x =>
                    Player.Level >= 90
                        ? PartnerPriority.Job090Prio.GetValueOrDefault(
                            x.ClassJob.RowId, int.MaxValue)
                        : int.MaxValue)
                .ThenBy(x =>
                    Player.Level >= 100
                        ? PartnerPriority.Job100Prio.GetValueOrDefault(
                            x.ClassJob.RowId, int.MaxValue)
                        : int.MaxValue)
                .ToList();

            newBestPartner = filter.First();
```

And while I am no expert, it seems like it's causing the Job090Prio to take precedence over the Job100Prio even at Lv100. Could be wrong but either way, it's definitely not using the correct Lv100 priority list when doing Lv100 content.

zbee:
- [X] Make the two sets of partner priorities not conflict
closes #619
- [X] Early-Return when there is only one partner (instead of sorting that one player)

### DIFF
--- a/WrathCombo/Combos/PvE/DNC/DNC_Helper.cs
+++ b/WrathCombo/Combos/PvE/DNC/DNC_Helper.cs
@@ -364,6 +364,12 @@ internal partial class DNC
             // If it's the last step and there are no matches found, bail
             if (filter.Count == 0)
                 return false;
+            // If there's only one match, return it
+            if (filter.Count == 1)
+            {
+                newBestPartner = filter.First();
+                return true;
+            }
 
             var orderedFilter = filter
                 .OrderBy(x =>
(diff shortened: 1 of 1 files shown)


Kagekazu (approved):

zbee:
Moved to #608

PR #623, opened 2025-06-08
[DRK] Updated Blood ability descriptions, various re-organizations and minor improvements

zbee:
- [X] Updated Blood ability descriptions for 7.0
- [X] Add `Dark Mind` to the AoE in-rotation mitigation Feature, per 7.1 adding physical mitigation to it.
- [X] Outside of combat all Combos will now return the correct action in the basic combo instead of just the Replaced Action.
- [X] Reorganized Helper file parts
- [X] Reorganized my generalized action logic into its own file, instead of taking up most of the Helper file
- [X] Various minor improvements to consistency of method usage
- [X] Various minor changes to throttling

> Only the first three items have any rotation adjustment, logic, or user-facing changes whatsoever.


Kagekazu (approved):
Worked fine on normal content.
found no issues in the new descriptions

edewen (approved):
I don't know DRK as well as other things But a couple of things stood out. Built it. it works thought so not game breaking. 

Rampart is in the aoe section, but not the single target section. Possible oversight?

Dark mind is in neither mit sections. as it is no longer just magica damage and is at least a 10% mit, again unsure for the reason for it being omitted.

When out of combat it drops back to hard slash. Extremely minor, but if you are still mid combo, that mid combo could be carried to the next target instead of forcing the combo restart.

PR #626, opened 2025-06-08
[UI] Fix Sheet Display Error for System Actions

Kaeris-Tempest:
- `System` actions have a `UnlockLink.RowId` value of `1` but no quest associated with them, resulting in an error.


zbee (approved):
Thanks for the fix :)

PR #627, opened 2025-06-08
[AST] Additions And Tweaks

edewen:
- [ ] Added the status overcap code to combust list
- [ ] Added Lightspeed Holding Option to retain one Lightspeed charge for burst
- [ ] Added Lightspeed Burst Option to Use aforementioned Lightspeed charge up to 2 gcds before divination
- [ ] Add Neutral Sect, going for a copy of scholars interaction between fey illumination and Succor.
- [ ] Add Sun sign
- [ ] Add Macrocosmos to aoe non bosses vs boss content.
- [ ] Add Sliders for Aspected Benefire, top end and bottom end.

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -225,6 +225,10 @@ public enum CustomComboPreset
     [CustomComboInfo("Lightspeed Weave Option", "Adds Lightspeed when moving", AST.JobID)]
     AST_DPS_LightSpeed = 1020,
 
+    [ParentCombo(AST_DPS_LightSpeed)]
+    [CustomComboInfo("Lightspeed Hold Option", "Retains 1 Lightspeed charge for use in the burst window", AST.JobID)]
+    AST_DPS_LightSpeedHold = 1061,
+
     [ParentCombo(AST_ST_DPS)]
     [CustomComboInfo("Lucid Dreaming Weave Option", "Adds Lucid Dreaming when MP drops below slider value", AST.JobID)]
     AST_DPS_Lucid = 1008,
@@ -232,6 +236,10 @@ public enum CustomComboPreset
     [ParentCombo(AST_ST_DPS)]
     [CustomComboInfo("Divination Weave Option", "Adds Divination", AST.JobID)]
     AST_DPS_Divination = 1016,
+
+    [ParentCombo(AST_DPS_Divination)]
+    [CustomComboInfo("Lightspeed Burst Option", "Add Lightspeed before Divination", AST.JobID)]
+    AST_DPS_LightspeedBurst = 1064,
 
     [ParentCombo(AST_ST_DPS)]
     [CustomComboInfo("Card Draw Weave Option", "Draws your cards", AST.JobID)]
@@ -278,6 +286,10 @@ public enum CustomComboPreset
     [CustomComboInfo("Lightspeed Weave Option", "Adds Lightspeed when moving", AST.JobID)]
     AST_AOE_LightSpeed = 1048,
 
+    [ParentCombo(AST_AOE_LightSpeed)]
+    [CustomComboInfo("Lightspeed Hold Option", "Retains 1 Lightspeed charge for use in the burst window", AST.JobID)]
+    AST_AOE_LightSpeedHold = 1062,
+
     [ParentCombo(AST_AOE_DPS)]
     [CustomComboInfo("Lucid Dreaming Weave Option", "Adds Lucid Dreaming when MP drops below slider value", AST.JobID
     )]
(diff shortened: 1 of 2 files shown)


WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -225,6 +225,10 @@ public enum CustomComboPreset
     [CustomComboInfo("Lightspeed Weave Option", "Adds Lightspeed when moving", AST.JobID)]
     AST_DPS_LightSpeed = 1020,
 
+    [ParentCombo(AST_DPS_LightSpeed)]
+    [CustomComboInfo("Lightspeed Hold Option", "Retains 1 Lightspeed charge for use in the burst window", AST.JobID)]
+    AST_DPS_LightSpeedHold = 1061,
+
```
zbee, line 231: Should this really be separate from `AST_DPS_LightspeedBurst`? I feel like they are intrinsically related.

WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -447,7 +475,7 @@ public enum CustomComboPreset
 
     #endregion
 
-    // Last value = 1060
+    // Last value = 1065
```
zbee, line 478: `1067`

WrathCombo/Combos/PvE/AST/AST.cs
```diff
@@ -263,37 +306,47 @@ protected override uint Invoke(uint actionID)
...
+
 
             if (!LevelChecked(AspectedHelios)) //Level check to return helios immediately below 40
                 return Helios;
 
             if (IsEnabled(CustomComboPreset.AST_AoE_SimpleHeals_LazyLady) &&
                 ActionReady(MinorArcana) &&
                 Gauge.DrawnCrownCard is CardType.Lady
-                && canLady)
+                && weaveLady && healthLady)
```
zbee, line 325: I'd personally suggest adding verbs to these variable names, `CanWeaveLady`, `HasHealthLady`, etc, but that's entirely up to you.

WrathCombo/Combos/PvE/AST/AST_Config.cs
```diff
@@ -115,6 +121,17 @@ internal static void Draw(CustomComboPreset preset)
                     DrawAdditionalBoolChoice(AST_AOE_DPS_OverwriteCards, "Overwrite Non-DPS Cards", "Will draw even if you have healing cards remaining.");
                     break;
 
+                case CustomComboPreset.AST_AOE_DPS_MacroCosmos:
+
+                    DrawHorizontalRadioButton(AST_AOE_DPS_MacroCosmos_SubOption,
+                        "Non-boss Encounters Only", $"Will not use on bosses", 0);
+                    DrawHorizontalRadioButton(AST_AOE_DPS_MacroCosmos_SubOption,
+                        "All Content", $"Will use in all content", 1);
```
zbee, line 129: Good call

zbee (commented):
Just figured I would drop some info while I had it open, nothing really pressing; good progress

WrathCombo/Combos/CustomComboPreset.cs
edewen: It is connected to the movement section because it actually affects how the movement lightspeed works.  It only mentions burst so people would know what they are holding it for. 

The lightspeed burst option is the on that is actually separated and connecting to divination instead because it is set up to fire 5 seconds before divination comes up.

zbee: I mean, I think that makes sense, but I was just thinking aloud that the Burst option could just retain the 1 charge away from movement instead.
Currently if you have the second option but not the first ... there may well be no Lightspeed to use, despite having the option on, if moving a lot, right?

WrathCombo/Combos/PvE/AST/AST.cs
edewen: Verbs applied

WrathCombo/Combos/CustomComboPreset.cs
edewen: If Burst retained one charge on its own, then there is bound to be someone who wanted to retain a charge from movement and do their burst manually and they are sure to complain. 

as for the looks. the way it shows up on the UI shows them all closely connected which i think works well

image

zbee: Having them next to each other is definitely better.

> then there is bound to be someone who wanted to retain a charge from movement and do their burst manually and they are sure to complain.

As for this, yeah, maybe, but it clearly says it will do that, so that's not really our problem.
Option 20 should at least say that it needs Option 18 to guarantee that a charge will be available. What we will definitely get more reports for is "I turned on Lightspeed Burst Option, but it doesn't do it, despite saying it will".

Kagekazu (approved):
Tested in M8N without issues

Kagekazu (approved):
Worked fine for me in M8N

WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -394,6 +413,10 @@ public enum CustomComboPreset
     [CustomComboInfo("Horoscope Option", "Adds Horoscope.", AST.JobID)]
     AST_AoE_SimpleHeals_Horoscope = 1026,
 
+    [ParentCombo(AST_AoE_SimpleHeals_AspectedHelios)]
+    [CustomComboInfo("Neutral Sect Option", "Adds Neutral Sect and its followup Sun Sign.", AST.JobID)]
+    AST_AoE_SimpleHeals_NeutralSect =1067,
```
zbee, line 418: missed a space after the `=`.

zbee (approved):
Logic changes and addition all look completely sound; not tested in game.

WrathCombo/Combos/CustomComboPreset.cs
edewen: fixed

PR #631, opened 2025-06-09
[RDM] Death to Advanced Action ID Selection

Tartarga:
- Ripped out Advanced Action Items in Single Target and AoE Advanced, while creating Features/Options that should cover existing flexibility / functionality
- Rearranged UI, edited names and descriptions
- Removed Simple Jolt combo now things in Advanced is always on Jolt (Spell Combo)
- TryOGCDs made more flexible for multiple uses (simple/adv, etc)
- Created RDM oGCD UI Drawer for multiple uses
- Added Riposte oGCD Option
- Added Reprise oGCD Feature
- Added Veraero/verthunder Feature
- Bugfix: TryAcceleration updated to take parameters to make Acceleration AoE options work. Those UI elements are now wired up
- Bugfix: Separated Scorch/Resolution from finisher, leaving it in the hands of OriginalHook as both replace ST and AoE buttons normally.  Scorch/Resolution will always be available regardless if Verflare / Verholy are enabled (fixes edge case with Grand Impact report in Discord)
- Renamed MeleeFinisher to Verflare / Verholy option
- Riposte Combo uses existing helper functions
- Removed legacy ComboTimer checks from Melee Combos


Kagekazu (approved):
Havent found an issue so far 😃

edewen:
Coming from a user standpoint here. 2 issues Im noticing. 

1, there are small features in between advanced st and advanced aoe that i almost mistook for being part of advanced st. IM sure just about every class has ordered st simple/ aoe, then advance simple aoe at the top. 

image

2. more just a personal standpoint, users. are not that smart. Something learned with dancer, they will check every box in single target and wonder why their melee isnt working. expect a lot of reports as long as the logic on this stays. 

image

zbee (changes requested):
Agreed on Edewen's second point.
That should be inverted to "Include Riposte", or should show some sort of warning text, at least.

Tartarga:
Item 1: I concur, I'll move it
Item 2: It's a Catch-22.  Goal was to make RDM more Auto Rotation friendly, so the thought process was to include the entire combo first, then add this exception.  If I write it the other way around, it's not auto rotation friendly unless you check that second box.  We all know people don't read and it'll mess up things either direction.  Now I'm not opposed to light things up like a Christmas tree with warnings if that box is checked and autorotation is on. This is what i bugged you (@zbee) about auto rotation & preset detection a few days back

zbee (commented):
I expect to review more fully later today, but at present: Happy with the UI facelift, haven't tested in game.

edewen (approved):
Ty, was not looking forward to seeing people every day confused by turning off thier own melee without reading.

PR #634, opened 2025-06-10
[TANK] Retargetting Things

### LINKED ISSUE #47: [Tanks] Off-tanking Overcap Support
Specifically for paladin but It could apply to other tanks as well. 

When tanking, There is overcap protection using Sheltron. The idea is, when not the targets target, to instead of using sheltron for overcap, to use Intervention on the targets target assuming they are

1 a tank
2 not invuln
3 in range
4 not already under the same effect

This could definitely apply to the other tanks offtanking tools as well.

edewen:
- [x] PLD Retarget Clemency Feature. Mouseover > hard target > lowest hp
- [x] PLD Sheltron => Intervention Retarget Feature
- [x] Fix Warrior Bloodwhetting feature to be Raw intuition/Bloodwhetting to Nascent Flash Feature, Correct logic. 
- [x] zbee added oblation logic for waste prevention
- [x] GNB Upgraded Aurora protection to include retargeting choices which also have the same built in protection. 
- [x] GNB Retarget Heart of Corundum

image
image
image
image

closes #47

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -2746,8 +2746,21 @@ public enum CustomComboPreset
     #endregion
 
     #region Aurora Protection
+    [ReplaceSkill(GNB.Aurora)]
     [CustomComboInfo("Aurora Protection Feature", "Locks out Aurora if Aurora's effect is on the target by replacing it with Savage Blade.", GNB.JobID)]
     GNB_AuroraProtection = 7023,
+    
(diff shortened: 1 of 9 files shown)


Kagekazu (approved):
Tested PLD and WAR without issues.
Haven't had time for the other 2 tanks.

Codewise it looks good

zbee (approved):
Tested half in game, and helped to debug/implement this in the first place; all looks sound to me otherwise.

PR #636, opened 2025-06-10
[DRK] Fix Opener sending again if failed very early

zbee:


Kagekazu (approved):
LGTM

Haven't tested in game

edewen (approved):
Logic seems sounds to prevent opener after beginning.

PR #637, opened 2025-06-12
[Core/Presets] Added Attribute to always show the User Options for a preset

Tartarga:
Current Behavior, user options are hidden when preset is not enabled
image

Depending on the situation (RDM / RPR Soulsow Reminder), it might be nice to go ahead and show the options regardless. Pictures taken with RDM PR. 

image
image
image

Using the attribute tag on Reaper's Soulsow Reminder would be another example
image
image

### DIFF
--- a/WrathCombo/Attributes/AlwaysShowUserOptionsAttribute.cs
+++ b/WrathCombo/Attributes/AlwaysShowUserOptionsAttribute.cs
@@ -1,0 +1,10 @@
+﻿using System;
+
+namespace WrathCombo.Attributes
+{
+    /// <summary> Attribute forcing User Options to always show, regardless of preset being enabled </summary>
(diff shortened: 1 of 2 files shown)


Taurenkey:
This should only happen if the user isn't hiding subcombo options as per the setting. Reason being hiding irrelevant things helps communicate that they shouldn't be in effect, so if this is always shown it *could* communicate that even though the main option is disabled, the bits under it are still applying.

zbee (changes requested):
I'll reiterate what Tauren said as an actual request for change, this should be based on the "Hide Children" Setting that already exists, as we talked about as well.
I'll write up a PR to your branch later tomorrow to do similar.

zbee:
It was a smaller change than I had in my head, so I just committed it directly. Re-request your review if you've nothing further to add, and I'll approve.

PR #640, opened 2025-06-14
[Jobs] Weekly stuffs

### LINKED ISSUE #628: [DRG] Recent Update Tries to Spam Mirage Dive
Title.

The most recent update tries to use Mirage Dive when High Jump is still on cooldown while using advanced mode ... and also when High Jump is _not_ on cooldown, rather than using High Jump itself. Please revert whatever change was made to DRG prior to the most recent update.

### LINKED ISSUE #638: DRG auto rotation stops after power surge is applied
I updated to the lastest version and noticed my auto rotation (and the non auto one) stops after power surge is applied when doing low level dungeons. (where I noticed it). If I remove the buff, the rotation starts again and then bugs out at power surge.

Upon looking more into it. Seems like it is only affecting simple rotation and I see mirage on the bar. I tried advanced with no added options and things are going fine.

Kagekazu:
DRG:
- Mirage dive fixes
- Cleanup of unneeded originalhook checks
- split jump/high jump so its easier to maintain
- cleanup of AoE for jumps and dives
- Fixes #628 
- Fixes #638 
- Removed Bosschecks in Simple Mode

MNK:
- Force `Riddle of Fire` to weave together with `Brotherhood`
- Removed Bosschecks in Simple Mode

RPR;
- Removed Bosschecks in Simple Mode

VPR:
- Removed Bosschecks in Simple Mode

### DIFF
--- a/WrathCombo/Combos/PvE/BLM/BLM.cs
+++ b/WrathCombo/Combos/PvE/BLM/BLM.cs
@@ -8,7 +8,7 @@ internal partial class BLM : Caster
 {
     internal class BLM_ST_SimpleMode : CustomCombo
     {
-        protected internal override CustomComboPreset Preset { get; } = CustomComboPreset.BLM_ST_SimpleMode;
+        protected internal override CustomComboPreset Preset => CustomComboPreset.BLM_ST_SimpleMode;
 
         protected override uint Invoke(uint actionID)
         {
@@ -175,7 +175,7 @@ internal partial class BLM : Caster
 
     internal class BLM_ST_AdvancedMode : CustomCombo
     {
-        protected internal override CustomComboPreset Preset { get; } = CustomComboPreset.BLM_ST_AdvancedMode;
+        protected internal override CustomComboPreset Preset => CustomComboPreset.BLM_ST_AdvancedMode;
 
         protected override uint Invoke(uint actionID)
         {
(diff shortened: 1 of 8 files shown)


edewen (approved):
Still lgtm after git did git things

WrathCombo/Combos/PvE/BLM/BLM.cs
```diff
@@ -8,7 +8,7 @@ internal partial class BLM : Caster
 {
     internal class BLM_ST_SimpleMode : CustomCombo
     {
-        protected internal override CustomComboPreset Preset { get; } = CustomComboPreset.BLM_ST_SimpleMode;
+        protected internal override CustomComboPreset Preset => CustomComboPreset.BLM_ST_SimpleMode;
```
zbee, line 11: Nice. I finally clicked on these and realized I could do this too when I was in DRK ha

WrathCombo/Combos/PvE/DRG/DRG.cs
```diff
@@ -258,11 +257,11 @@ protected override uint Invoke(uint actionID)
                         ActionReady(MirageDive) &&
```
zbee, line 257: If I recall the conversation about this correctly, you may also want to `OriginalHook(MirageDive) is MirageDive`

zbee (approved):
Logic and `()` look sound and make sense.
Jump Logic looks a lot cleaner.
Haven't tested in-game.

WrathCombo/Combos/PvE/DRG/DRG.cs
Kagekazu: Oh ye, I'll add it in when I'm back home! Forgot about that one

Kagekazu: > If I recall the conversation about this correctly, you may also want to `OriginalHook(MirageDive) is MirageDive`

This should be `OriginalHook(Jump) is MirageDive` right?

Taurenkey (approved):

PR #641, opened 2025-06-15
[AST] Lucid Fix

edewen:
Lucid was never wired up for aoe, it was using st config.


zbee (approved):
This will also fix the double "Reset to Default", @Kagekazu

zbee (approved):
Only looked at WHM and AST. They both look solid, especially the upper/lower regen limit. 👍🏻

Kagekazu (approved):
LGTM

PR #647, opened 2025-06-18
[WAR/Core] `Stationary Only` Delay Check + small `DrawSliderFloat` param for decimal places

Akechi-kun:
Since the values we're working with typically don't require precision beyond the first decimal place, supporting flexibility from one to three should be okay here and for other use-cases with other jobs. 

The default param set (3) will keep behavior as is for the other references, and now devs can set `decimals: 1` to reflect this behavior here in the screenshots with the slider moving in increments of `0.1` instead of `0.001`. 

image
image

I will admit this feels kind of lazy, or is this just really that simple? idk, but this works fine from testing😄

Let me know if there's a better way you'd like this done.

### DIFF
--- a/WrathCombo/Combos/PvE/WAR/WAR.cs
+++ b/WrathCombo/Combos/PvE/WAR/WAR.cs
@@ -158,7 +158,8 @@ internal partial class WAR : Tank
                 return Upheaval;
             if (IsEnabled(CustomComboPreset.WAR_ST_PrimalWrath) && ShouldUsePrimalWrath)
                 return PrimalWrath;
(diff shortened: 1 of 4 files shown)


Akechi-kun:
> Its set up for pooling in IR. But if you dont have IR enabled. It keeps pooling. Just something to look at. Works though.

@edewen go crazy king

zbee (approved):
I would rather see this config as a global Setting to apply to all Stationary checks, and for a `IsStationary()` method to be added to benefit from that setting, but that can be done later, and this is certainly a welcome addition.
I'll add it further down my todo list, if no one else gets to it.

PR #660, opened 2025-06-23
[SMN PVP] Tweaks for burst, and prevent r4 waste

edewen:
Shifted a couple things around to allow for a more frontloaded burst. And less chance to waste an r4 proc. 

Start with 2 Ogcds which sucks but then it will knock out 30k in 2 gcds followed by another 32k in 3 gcds. 

Added a line to throw up savage blade if target is immune during an r4 proc. It replaces ruin 3, so it is the only way to prevent a wasted proc.


zbee (approved):
lgtm

PR #662, opened 2025-06-24
[SGE] DoT fix at low FPS

Tartarga:
Fixes edge case where _DosisList[OriginalHook(Dosis)])_ in DoT refresh code could return "Eukrasian Dosis" due to limited/potato FPS. Added Eukrasian Dosis set to DosisList as the solution.  Converted to new FrozenDictionary for faster lookups.


edewen:
I believe Kage put all this in his pr

zbee (approved):

Tartarga:
Implemented in #663

PR #665, opened 2025-06-24
[SCH] WellReadMage Overhaul

edewen:
- General
  - Cleanup and regions for all sch stuff
  - Removed redundant options
  - Reorganized CCP to be more logical
 
- ST
  - Rewrote Dots in whm style
  - Split Chain Stratagem and Baneful Impaction into 2 options
 
- AoE
  - Added Chain Stratagem with a optional Level Check for Baneful Impaction
  - Added Baneful Impaction
  - Added Energy Drain
 
- ST Heals
  - Added Adloquim to the ST priority System
  - Added Emergency Tactics as a option on Adloquim with a separate health slider
  - Added Fey Blessing, Illumination, and Whispering dawn to ST with Non Boss Option for dungeon trash (single target options   were lacking there and was using gcds to heal too often with autorotation)
  - Added the Dissolve union to all 4 combos. Will not affect dps, but it is needed for auto rotation to turn off Aetherpact when the target reaching the desired health level. This is a bug fix for already existing code. 
 
- AoE Heals
  - Fix Dissipation Logic, if Aetherflow was held for Indomitably, Dissipation would just fire instead.
  - Added hold for Indomitably option to Dissipation as well.
  - Added Summon Seraph to the priority List
  - Added Emergency tactics option to Succor, similar to ST. It will use ET if the Shield check fails and no other priorities exist 
  - Fixed issue for 20-35. Succor level check, whispering dawn only option at that level range.
 
- Standalone Features
  - Added Sacred Soil Retarget Standalone
 
- Hidden Features
  - Added Hidden Raid Wide Features for Expedient, Sacred Soil, and Succor as raidwide still needs a lot more testing before public access.

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -6135,29 +6135,41 @@ public enum CustomComboPreset
     SCH_ST_Heal_Esuna = 16026,
 
     [ParentCombo(SCH_ST_Heal)]
-    [CustomComboInfo("Adloquium Option", "Use Adloquium when missing Galvanize or target HP%% below:", SCH.JobID)]
+    [CustomComboInfo("Adloquium Option", "Use Adloquium", SCH.JobID)]
     [PossiblyRetargeted]
     SCH_ST_Heal_Adloquium = 16027,
 
     [ParentCombo(SCH_ST_Heal)]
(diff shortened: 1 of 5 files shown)


zbee (approved):
The logic for everything mentioned in the description looks good and sound to me; haven't tested in game.

PR #668, opened 2025-06-30
[Melees] Some updates 

Kagekazu:
- MNK
  - Add Hidden feature for M6S

- SAM
  - Cleanup of Obselete `TargetIsHostile`

- DRG
  - Add choice for movement and in melee checks

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -3781,6 +3781,8 @@ public enum CustomComboPreset
     MNK_Variant_Cure = 9026,
 
     #endregion
+    
+    #region Misc
 
     [ReplaceSkill(MNK.PerfectBalance)]
     [ConflictingCombos(MNK_PerfectBalanceProtection)]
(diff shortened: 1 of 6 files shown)


Kaeris-Tempest:
Did a bit of streamlining
- Converted the lists into HashSets so we can use `.Contains` for instant `O(1)` lookups
- Simplified the conditionals so it all flows off of `remainingGCD`

WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -5714,9 +5727,31 @@ public enum CustomComboPreset
...
+    
+    [ParentCombo(SGE_Hidden)]
+    [CustomComboInfo("Holos Option", "Will try to cast Holos when a raidwide casting is detected. \nWill be used in all 4 main combos.", SGE.JobID)]
+    [Hidden]
+    SGE_Hidden_Holos = 14072,
+    
     #endregion
 
-    // Last used number = 14064
+    // Last used number = 14068
```
edewen, line 5754: My fault, but this should be 14072

Kagekazu: Ye I'll add the correct one when I move the openers

zbee (approved):
Changes look sound, thanks for clearing the warnings; didn't test in game, besides PLD's minor change.

PR #669, opened 2025-06-30
[AST] Priority Treatment

edewen:
- General
 - A bit of general cleanup utilizing helper file for the more complicated looking card stuff. 
 - Organize it to be ST, AoE, ST Heal, AoE Heal, Standalones
 
- ST Heal Priority System
 - 11 Total Options
 - The 8 normal single target heals
  - Aspected Heal, Essential Dignity, Celestial Intersection, Exaltation, 4 healing cards
 - Added 3 AoE OGCDs, with Not on Bosses option to allow for better usage during dungeon trash. 
  - Celestial Opposition, Lady card, Collective Unconscious 
  
- AoE Heal Priority System
 - 9 Total Options
  - Aspected Helios and Helios, so the users choice of Alt Mode wont hinder Automatic. 
  - Horoscope and Horoscope Heal Separately
   - Horoscope Is set up to be followed directly by a form of Helios to enabled Horoscope Heal
  - Celestial Opposition, Collective Unconscious, Lazy Lady, Neutral Sect
  - Stellar Detonation only when you have reached giant dominance

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -821,8 +821,11 @@ public enum CustomComboPreset
     [ParentCombo(AST_AoE_SimpleHeals)]
     [CustomComboInfo("Stellar Detonation Option", "Adds Stellar Detonation when under the effect of Giant Dominance", AST.JobID)]
     AST_AoE_SimpleHeals_StellarDetonation = 1072,
-
     
+    [ParentCombo(AST_AoE_SimpleHeals)]
+    [CustomComboInfo("Collective Unconscious Option", "Adds Collective Unconscious for to regen, it will not hold it.", AST.JobID)]
+    [PossiblyRetargeted]
+    AST_AoE_SimpleHeals_CollectiveUnconscious = 1074,
 
     [ReplaceSkill(AST.Benefic2)]
     [CustomComboInfo("Benefic 2 Downgrade", "Changes Benefic 2 to Benefic when Benefic 2 is not unlocked or available.",
@@ -875,7 +878,7 @@ public enum CustomComboPreset
 
     #endregion
 
-    // Last value = 1072
+    // Last value = 1074
 
     #endregion
 
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -719,6 +719,16 @@ public enum CustomComboPreset
         AST.JobID)]
     [PossiblyRetargeted]
     AST_ST_SimpleHeals = 1023,
+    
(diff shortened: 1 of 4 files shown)


WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -765,34 +771,67 @@ public enum CustomComboPreset
...
     [PossiblyRetargeted]
     AST_ST_SimpleHeals_Bole = 1050,
+    
+    [ParentCombo(AST_ST_SimpleHeals)]
+    [CustomComboInfo("Celestial Opposition Option", "Adds Celestial Opposition", AST.JobID)]
+    [PossiblyRetargeted]
+    AST_ST_SimpleHeals_CelestialOpposition = 1068,
+    
+    [ParentCombo(AST_ST_SimpleHeals)]
+    [CustomComboInfo("Collective Unconscious Option", "Adds Collective Unconscious for to regen, it will not hold it.", AST.JobID)]
```
zbee, line 781: > "for to regen"

I think the model from the cards, at least, would be `Adds Collective Unconscious (for the Regen)`

WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -843,9 +882,34 @@ public enum CustomComboPreset
...
+    #endregion
+    
+    #region Hidden Features
+    [CustomComboInfo("Hidden Options", "Collection of cheeky or encounter-specific extra options only available to those in the know." +
+                                       "\nDo not expect these options to be maintained, or even kept, after they are no longer Current.", AST.JobID)]
+    [Hidden]
+    AST_Hidden = 1075,
+    
+    [ParentCombo(AST_Hidden)]
+    [CustomComboInfo("RaidWide Collective Unconscious Option", "Will try to Weave Collective Unconscious when a raidwide casting. \nWill be used in all 4 main combos.", AST.JobID)]
```
zbee, line 894: Could use some clarification here. Does this make `Collective Unconscious` ALSO used vs raidwides, or does it add a raidwide requirement for ALL of its usages across the combos?

zbee (changes requested):
Some minor things that I noticed while setting up the job.

WrathCombo/Combos/CustomComboPreset.cs
edewen: Done
"Adds Collective Unconscious (for the regen so it will not channel)" I have to leave the channel part or people will expect it to channel

WrathCombo/Combos/CustomComboPreset.cs
edewen: added the additionally for you

PR #671, opened 2025-07-01
docs for holding burst macros

nightraidwiki:
A list of macro for Tank & Dps to Hold Cooldown/Burst

### DIFF
--- a/docs/BurstHoldingMacros.md
+++ b/docs/BurstHoldingMacros.md
@@ -1,9 +1,13 @@
 # Wrath Macro Holding Burst
 
-You just have to create a macro with the line `/wrath toggle <id_of_the_setting>` to deactivate/activate a function in your wrath rotation.  
-It's important to do it for fights when you need to delay your burst.
+Per [Issue #99](https://github.com/PunishXIV/WrathCombo/issues/99), Burst Holding is still something we plan to implement more properly at a later time.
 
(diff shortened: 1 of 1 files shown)


edewen:
Noticed you didnt add healers yet. These should work for them if you wanna add

WHM Just turn off Presence of mind
/wrath toggle 19008
/wrath toggle 19195

SCH Toggling off chain strat, if everything enabled, energy draon shouldnt dump while it is off cd. 
/wrath toggle 16003
/wrath toggle 16054

AST toggles off div, card pooling is designed to wait for div buff to happen, so dont need to toggle the cards. 
/wrath toggle 1043
/wrath toggle 1016

SGE toggle off psyche and phlegma, needs both bc phlegma is coded that ti can happen before or after phyche based on its cd. 
/wrath toggle 14051
/wrath toggle 14010
/wrath toggle 14008
/wrath toggle 14005

zbee:
As Edewen said, burst holding is something we intend to do much differently, and these macros give people more control over what exactly gets toggled.

It would be fairly simple to add `/wrath toggle burst`, yes, but that would then be something we'd have to tweak and adjust based on user requests, when this is just a stop-gap thing, so that wouldn't make a lot of sense when we want our effort to go towards an actual burst holding option.

edewen (approved):

PR #673, opened 2025-07-03
[ActionWatching] Class Update

Kaeris-Tempest:
# TL:DR
- A bunch of tasks are now completed in a fraction of their original time.
- A bunch of places now use frozen collections for faster lookup speeds.
- Plugin window will no longer hammer the system with locale requests.
- Cleaned a bunch of defunct code; improved a bunch of functions.

### ReceiveActionEffectDetour
- Memory access **significantly** improved when processing effects `64 → 8`.
- Partylist access **significantly** improved when processing targets `64 → 8`.
- Added caching and referencing for several fields and properties.
- Debug logging improved:
```
[ActionEffect] Type: Damage | Value: 1500 | Params: [3, 6, 0, 0, 160] | Action: Fire IV (ID: 3577) → Target: Training Dummy | Flags: [AtSource: False, FromTarget: True]
```

### New List: WeaveActions
- Contains only abilities used between GCDs; gets cleared everytime a GCD is used.
- Provides finer control over the most recent weave window. Examples:
```
public static bool HasWeaved() => WeaveActions.Count > 0;
public static bool HasDoubleWeaved() => WeaveActions.Count > 1;
public static bool HasWeavedStardiver() => WeaveActions.Contains(DRG.Stardiver);
```

### SendActionDetour
- Added caching and referencing for several fields and properties.
- Debug logging improved:
```
[ActionSend] Action: Glare IV (ID: 25859) | Type: 1 | Sequence: 42318 | Target: Training Dummy | Params: [0, 0, 0, 0, 0]
```

### CustomComboInfoAttribute
- Cleaned all unused code, likely resulted from a partial migration.
- Streamlined and consolidated all `ClassJob` functions under `Misc.cs`.
- Added `TextInfo` cache to prevent excess locale requests when writing job names.


zbee (approved):
I do see a very minor fps gain (~1 at idle, maybe 2-3 across a dungeon), and I've not had any issues thusfar with this tonight at all.
With Edewen and Kage both mentioning (at least in Discord) that they've not had problems with it either: approved.

I'd be happy to dig in more, if it were held back a week, but considering the collaboration that we (you and I, but also others) have done on different parts of this I don't think that that is necessary.

Kagekazu (approved):
Haven't had any issues in the last few hours either. So it's good to go for me

PR #676, opened 2025-07-04
[SGE/SCH] Dash of cleanups

Tartarga:
SGE
- SGE_ST_DPS: Combined actionID check and replacedActions creation a single property, DosisActions
- SGE_Eukrasia typecast to Value edit
- Combined DosisList and DosisToEDosisList using touple SCH
- SCH_Recitation: Changed switch statement to expression, typecast to Value edit


edewen:
SCH_Recitation: Changed switch statement to expression, typecast to Value edit

Not really sure the why behind it. Assuming it is supposed to run better i guess? Code level above my head. but built and sch thingy works so this is me smiling and nodding

Kagekazu (approved):

PR #680, opened 2025-07-06
[BLM] make triplecast suboptions

Kagekazu:
- Add triplecast suboptions to use always or not when having leylines


zbee (approved):
Thanks.
I'd enable the flag to have the descriptions be shown as tooltips so I'd have the extra space to add a line like "this is the recommended behavior" to the end of the "not under leylines" description; but I'll leave that up to your discretion.

edewen (approved):

PR #681, opened 2025-07-06
[Docs] Add documentation into the `.sln` file

zbee:


Kaeris-Tempest (approved):
If Bee thinks its good, then it's good enough for me

PR #689, opened 2025-07-07
[Action] Weave Updates

Kaeris-Tempest:
Heavy WIP.

PR #693, opened 2025-07-12
[BLU] MoonFlute Fix

Kichijotenn:
Added a `!JustUsed(MoonFlute)` to all instances of MoonFlute on lvl 70 and 80 combos to avoid double casting it due to latency. @zbee


zbee:
@Kaeris-Tempest and I make tweak this after we rework `LastAction` stuff in #689, just btw

zbee (approved):
Easy fix, thank you

PR #703, opened 2025-07-17
[IPC] Prevent network checks from being problematic

zbee:
- [X] Only check every 45 minutes instead of 20
- [X] Check on a side thread


edewen (approved):
Looks like it makes sense to me.

PR #706, opened 2025-07-19
[UI] Update Sidebar's default size, +min size, +collapse button

zbee:
- [X] Changes the default size of the plugin's Left Column (the column with the tabs and icon)
      It is now the width of the plugin icon, plus some, instead of 1/3 the window size.
- [X] Adds a minimum size to the Left Column
      (which is the plugin icon's width)
      If the user reduces the size below the minimum, the column sets itself to be non-resizable (to stop the drag event), and sets itself to the default column width.
- [X] Adds a collapse button, at the bottom left
      To prevent the loss of being able to crunch the sidebar down to its minimum size (which was never zero) I have also added a collapse button.
      When pressed, it redraws the window to a table with a singular column for only the body of the content, removing the sidebar.
      (note: this also changes the ID of the table, to prevent the sidebar being set to the full width of the window when un-collapsed after being collapsed)

> The code diff is not particularly readable on GitHub nor SemanticDiff, because the code needed broken into different methods to support conditional rendering.
> I would recommend just pulling this PR and trying it instead of trying to look at these diffs.

### DIFF
--- a/WrathCombo/Core/PluginConfiguration.cs
+++ b/WrathCombo/Core/PluginConfiguration.cs
@@ -128,6 +128,8 @@ namespace WrathCombo.Core
         public bool ShowHiddenFeatures = false;
 
         public bool SuppressQueuedActions = true;
+        
+        public bool UILeftColumnCollapsed = false;
 
         #endregion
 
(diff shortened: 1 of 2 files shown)


edewen (approved):
Built. LGTM.

Kagekazu (approved):
Tried in game with no issues

PR #707, opened 2025-07-19
[AST] Early Stellar Detonation Option + dancer tiny fix for level check

edewen:
image

Added option for early stellar detonation, with boss toggle and hp slider to help prevent waste on dungeon trash. Properly set for non bosses would not interfere with the party hp check for aoe healing that also exists. At least how I would use it. 

did a quick fix for dancer issue not using fans at lower levels. it was being blocked in st from it trying to pick a dance partner before it was high enough to use closed position


Kagekazu (approved):
LGTM

zbee (approved):
Built, tested in-game, Detonation looks great and works, and all UI for it works. Thanks for the DNC fix.

PR #708, opened 2025-07-20
[AutoRot] Basic Invulnerability checking for Nier 3

Tartarga:
Added basic logic for The Tower at Paradigm's Breach, Hansel & Gretel. AutoRot will treat parrying boss(s) as invulnerable. Tanks will have a directional check


edewen:
Tested as warrior. Tanked Hansel.

When hansel has shield. it used abilities just fine. 

But when the other boss had the shield and hansel had the spear. It stopped attacking.

edewen:
Simplified the code back down as it did not work. As the bosses don't always have buffs, targeting is not exactly a concern of the ignore code. if tank, be in front of your target.

edewen (approved):

zbee (approved):
looks much better now, untested by me, but known to be tested for tank and non-tank by y'all, so it's got my confidence in that regard too.

PR #709, opened 2025-07-20
[BLM] remove Originalhook from Paradox

Kagekazu:


Kagekazu:
will do it next update

PR #712, opened 2025-07-22
[AutoRot] Fix error spam from `Esuna`

zbee:
- [X] Fixes `Esuna` in Auto Rotation not checking if the target can receive `Esuna`


edewen (approved):

Kagekazu (approved):
Tested esuna part in game,
Haven't checked about the jobstone, but code seems to be good.

PR #715, opened 2025-07-23
[Docs] Auto rotation and Healer Setup Guides

edewen:


zbee (approved):
Looks good, healer settings will be welcome.
Excited to cannibalize both of these files into the Help tab one day.

PR #718, opened 2025-07-26
[SAM] Add bosscheck meikyo usage

Kagekazu:
Add bosscheck to meikyo usage.

Makes it so that u can choose wether to always use the balance logic, or not use it in non boss content.
(its like Queen logic of MCH)


edewen (approved):
logic lgtm. im no sam though

PR #724, opened 2025-07-30
[Various] General code cleanup/BLM - MNK - RPR Updates

### LINKED ISSUE #726: [BLM] AoE rotation bug is back
Noticed it once before, seems its back again. Freeze/B2 does not appear in AoE rotation when below level 60 to recover MP. Hopefully my debug file will provide some help.

[WrathDebug.txt](https://github.com/user-attachments/files/21563969/WrathDebug.txt)

Kagekazu:
BLM

- Update Standalone options
  - Update Fire I / Fire III
  - Update Blizzard 4 to Despair
    - Add option to choose B4 or B3
  - Fix Fire 1 to Despair
- Add needing a battle target for movement options
- Closes #726 

MNK
- Add in opener prepull.
- Add Thunderclap retargetting
- Condense `ROF- BH` and `BH - ROF` into 1 option where u can choose the button.
- Update Perfect Balance usage 
- Made sure Fire's reply is always inside buff windows
- update Wind's reply usage to be used in buffs when possible

RPR
- Update to follow new graph
image

Various
- cleanups of unneeded Parenthese
- Transformed most to switch or switch expressions

### DIFF
--- a/WrathCombo/Combos/PvE/MNK/MNK.cs
+++ b/WrathCombo/Combos/PvE/MNK/MNK.cs
@@ -513,8 +513,8 @@ internal partial class MNK : Melee
         protected internal override CustomComboPreset Preset => CustomComboPreset.MNK_Retarget_Thunderclap;
 
         protected override uint Invoke(uint actionID) =>
-            actionID is Thunderclap 
-                ? Thunderclap.Retarget(SimpleTarget.Stack.MouseOver ?? SimpleTarget.HardTarget, true) 
+            actionID is Thunderclap
+                ? Thunderclap.Retarget(SimpleTarget.Stack.MouseOver ?? SimpleTarget.HardTarget, true)
                 : actionID;
     }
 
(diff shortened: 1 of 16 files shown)


edewen (approved):
lgtm. not tested in game bc i dont know jack about blm yet but logic seems sound for things.

WrathCombo/Combos/PvE/MNK/MNK.cs
```diff
@@ -540,6 +503,41 @@ protected override uint Invoke(uint actionID)
         }
     }
 
+    internal class MNK_Retarget_Thunderclap : CustomCombo
+    {
+        protected internal override CustomComboPreset Preset => CustomComboPreset.MNK_Retarget_Thunderclap;
+
+        protected override uint Invoke(uint actionID) =>
+            actionID is Thunderclap
+                ? Thunderclap.Retarget(SimpleTarget.Stack.MouseOver ?? SimpleTarget.HardTarget, true)
```
zbee, line 512: I don't think you need to fall back to Hard Target here, that's vanilla behavior, and the retarget doesn't happen if the target is `null`.

PR #725, opened 2025-08-02
[Core] Auto Rotation pausing for some Acceleration Bombs

Tartarga:
Public Change:
Auto Rotation should now pause for Weeping City's Ozma & Vanguard's Protector Acceleration Bombs

Internal Changes
Sorted area list
Status Caches now Frozen Sets

### DIFF
--- a/WrathCombo/CustomCombo/Functions/Status.cs
+++ b/WrathCombo/CustomCombo/Functions/Status.cs
@@ -2,6 +2,7 @@
 using ECommons.DalamudServices;
 using ECommons.GameFunctions;
 using FFXIVClientStructs.FFXIV.Client.Game;
+using ImGuiNET;
 using System.Linq;
 using WrathCombo.Data;
 using WrathCombo.Services;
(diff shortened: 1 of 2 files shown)


edewen (approved):
Looks good enough to me. Not tested in game.

WrathCombo/CustomCombo/Functions/Status.cs
```diff
@@ -2,6 +2,7 @@
 using ECommons.DalamudServices;
 using ECommons.GameFunctions;
 using FFXIVClientStructs.FFXIV.Client.Game;
+using ImGuiNET;
```
zbee, line 5: Looks like this is not actually needed

ECommons
```diff
@@ -1 +1 @@
-Subproject commit b3508ff1e6afbaf04d839b8815509513fd22ab2f
+Subproject commit 512f339c5137a226a24d6de9860efad63df56718
```
zbee, line 1: This also looks like it is unnecessary

zbee (changes requested):

Taurenkey:
Don't forget to unset the target since autoattacks could still trigger.

ECommons
Tartarga: I updated submodules and accidentally commited.  Will fix on rebase

Tartarga:
Closing in favor of #727

PR #727, opened 2025-08-03
[Core] Added Acceleration Bomb & Pyretics

Tartarga:
Public
Introduced PlayerHasActionPenalty. Checks at the moment for Acceleration Bombs and Pyretics. Will unselect current target upon detection. Needs to be wired up in appropriate locations

Internal
StatusCache had many hashsets converted to frozen, AccelerationBombs and Pyretics frozensets added. TargetIsInvincible map IDs sorted


edewen (approved):

zbee (approved):
looks good as is

as for the wiring up of `PlayerHasActionPenalty`:
We can either leave this open for me to add that here later, or I can just do another PR later, and this can be unused until then; either way works.

Tartarga:
Wire it up directly in this PR if you'd like
I wired it up temporarily to test by putting it in Debug and kept the debug window open at all times

zbee:
Added action-blocking to Auto-Rotation. Will have to be more clever another time to add it to manual. Updated to current `main`.

PR #732, opened 2025-08-07
[Core] Meso Terminal Second Boss Invulnerable check

edewen:
see title

all bosses come with the same assignment every time. Healers SEEM to always be alpha and tanks beta, and dps is either gamma or delta. 

added hellmaker priority when it lights your floor on fire as a dps


zbee (approved):
I am wary of your comment that healers "seem to" always follow this, but I assume that's something that depends on strategy (otherwise I'd hope for more testing).
These Data IDs and Status IDs seem to be correct though, and the logic looks sound though, so assuming my assumption is correct: approved and thanks.

edewen:
im wary of it too which is why i coded it so if the healers and tanks do swap at some point and i never saw it. the code would handle it fine

Tartarga (approved):

PR #734, opened 2025-08-08
[Core] Cleanup

Kagekazu:
- make everything file scoped namespaces
- remove unused `using`
- move to global imgui binding
- Move to global alias Preset = WrathCombo.Combos.CustomComboPreset;
- Applied static WrathCombo.Window.Functions.UserConfig; to every job config file missing it
- Applied job specific static members to each job/helper file missing it


WrathCombo.slnx
```diff
@@ -1,35 +1,26 @@
...
-    </Folder>
-
-    <Configurations>
-        <BuildType Name="Debug" />
-        <BuildType Name="Release" />
-        <Platform Name="x64" />
-    </Configurations>
-
-    <Project Path="ECommons/ECommons/ECommons.csproj">
-        <Platform Project="x64" />
```
zbee, line 27: Put these back please. Rider only thinks they're wrong.

WrathCombo/Combos/PvE/ALL/ALL.cs
```diff
@@ -5,9 +5,11 @@
 using WrathCombo.Extensions;
 namespace WrathCombo.Combos.PvE;
 
-internal partial class All
+internal partial class ALL
```
zbee, line 8: I think just a couple months ago we changed `ALL` to `All` ha

Kagekazu: think that's from @Tartarga 's PR

zbee: Yeah I do believe so

edewen (approved):
So much but I dont see anything funny since yesterday when i had it built.

PR #735, opened 2025-08-08
[BRD] Multidotting

edewen:
- New Feature: Bard Multidotting. Both ST and AoE up to 3 targets as more is a dps loss
- Configs for buffs and dots have changed.
- Updated Burst Hold doc with new macro to turn off buffs.


zbee (approved):
Helped setup and fix the `SimpleTarget` for this, all logic looks good

Kagekazu (approved):
logic seems good, havent tested in game

PR #738, opened 2025-08-09
[BLM] updoots

Kagekazu:
- Add Target HP% Slider to avoid using Leylines.
- Add Retargeting Feature for Aetherial Manipulation.
- Add Time Player Standing Still Slider for LeyLines to skip LeyLines if you haven't been still for at least that long.
- Add Movement Options for AoE


zbee (approved):
This feels much better, thank you.

edewen (approved):

PR #743, opened 2025-08-11
[Various] some forgotten fixes

Kagekazu:
MNK

- Fix basic combo overwriting all 3 skills
- Make it so that basic combo uses configs instead of preset options

RPR
- Fix soulsow showing on all options

VPR
- Fix Reawaken standalone option 
- Fix Serpents Tail standalone option


edewen (approved):

zbee (approved):
Looks generally sound, and it appears that all Configs still line up with what was being checked before.

PR #746, opened 2025-08-11
[Core] Undo Auto-Rotation Targeting change

zbee:
- [X] Undoes #730, which was causing Auto-Rotation to not have a target to be used for (particularly ranged) Actions


Tartarga (approved):
Fixed my random issues in new dungeon. Approved

PR #750, opened 2025-08-12
Conversion to ECommons Job enum

Tartarga:
Public facing, some UI Changes, and abbreviations used in /wrath commands will compare to the sheets. so if client is non english and they use different abbreviations, those will be valid now

Replaced alot of jobID references to Job Enum
Removed functions in Misc.cs, relying on ecommons code Removed Dictionary'd ClassJob sheet
Removed Island Sanctuary Sprint
Frozen Dictionaries for Role groups now generate by the sheet data, not hard entries General changed to Roles and Content. Skeleton work done on Variant. Occult moved into it Redid Occult Job icon generation, storage, and display since generation no longer takes place on screen open

Will be submitting more Job requests to ECommons for further cleanup (abreviation, name, etc)


Tartarga:
Everything is done as of 2025-08-14, ready for review

Tartarga:
Made a mistake interpreting one of the /wrath commands. Setting PR back to draft for now.

Tartarga:
Setting back to draft, awaiting EC PR acceptance/denial for a fix

zbee (commented):
I've chipped through the code bit by bit over the past few days, and it all seems good, and every test I can think of has passed.
But I'm not particularly comfortable approving it for release, particularly since it only actually became ready yesterday.

I think this would be a good candidate to be merged after release, for us to play with all week long.

Kagekazu (approved):
Havent found any issues so far

zbee (approved):
Approved for after-1.0.2.3 merge.

PR #751, opened 2025-08-13
[NIN] Gonna be the Hokage

edewen:
- Overhaul of Ninja. ST and AoE
  - Many features Added, Adjusted, or Removed. Settings have been reset.
  - Expanded on the Hide Standalone. Instant Toggling Hidden off Option.
  - ! In Testing, learned that if you are getting manual press mudra fails, the reason is likely the action queue. One can use Reaction to lower the Action Queue from .5, to buy a tenth of a second or so allowing wrath to do its thing without any major affect on gameplay.
  
https://xivanalysis.com/fflogs/7jTGKgMFXc4Jt8NB/52/2
13 min parse. Zero suggestions.

### DIFF
--- a/WrathCombo/AutoRotation/AutoRotationController.cs
+++ b/WrathCombo/AutoRotation/AutoRotationController.cs
@@ -606,7 +606,9 @@ internal unsafe static class AutoRotationController
                     Service.ActionReplacer.getActionHook.IsEnabled ? gameAct : outAct,
                     (mustTarget && target != null) || switched ? target.GameObjectId : Player.Object.GameObjectId);
 
-                if (outAct is NIN.Ten or NIN.Chi or NIN.Jin or NIN.TenCombo or NIN.ChiCombo or NIN.JinCombo && ret)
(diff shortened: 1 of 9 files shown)


zbee (changes requested):
Simple Modes in Manual-use perform well, code looks good.

Feint should check if the target already has Feint and if Feint can be applied; probably via `RoleActions`.
(`RoleActions.Melee.CanFeint()` should be updated like `CanReprisal` was to check `CanApplyStatus`)

Much of NIN works off of debuffs and should check `CanApplyStatus` in some way.

zbee (approved):
Simple Modes in Manual-use perform well, code looks good.

Couple of notes that don't necessarily need done now:

- Feint should check if the target already has Feint and if Feint can be applied; probably via `RoleActions`.\
  (`RoleActions.Melee.CanFeint()` should be updated like `CanReprisal` was to check `CanApplyStatus`)
- `CanApplyStatus` should probably be sprinkled throughout NIN, since it uses a fair bit of debuffs.

Kagekazu (approved):
Advanced mode works fine in current content.
Tested it in latest dungeon.

i do recommend as zbee said to add some status checks in later.

edewen:
Ill get feint stuff added next week for sure

PR #763, opened 2025-08-18
[Melees] Add raidwide Feint + stun

Kagekazu:
- Add Advanced raidwide feint option
- Add Advanced ST and AoE stun option
- add target check for RPR, MNK and VPR burst


Kagekazu:
> Logic looks fine. But I personally would revisit them next week.
> 
> Your ogcd sections have all the mits and even interrupts at the highest priority ogcd.
> 
> True mit is important. but I know how touchy mch has been with mits in the past for you. having it dismantle or interupt during a touchy time in the rotation is terrible. Mits and interrupts should probably be down at the bottom with the selfheals.

moved them around.

Also reordered the combos in the file itself, so dont be alarmed about the diff, there is nothing actually changed except the order xD

PR #765, opened 2025-08-19
[Core] Variant

Tartarga:
Previously created "Variant Dungeons" tab under Roles and Content now completed
Did away with the variant interfaces/implementation complexity. Using the older approach with some new tricks (Player Role checking)
Removed individual class variant presets
Updated all classes to use new Variant calls
I'm Fired.


zbee:
Just curious about your commit message for 035bfd3:
Why not raises, too?

zbee:
As a guide, it could probably work Exactly like OC Chemist's `Revive`, if you wanted to nab that logic.

As for `Swiftcast` fighting, that's possible, but I wouldn't see it as a big issue, since Variant would be first, and that's the most "free" usage of the `Swiftcast` anyways, so that is likely ideal.

> Could also code it out for dead party member detection, prompt swiftcast then raise, but i need to look for that bit of code

`if (SimpleTarget.AnyDeadPartyMember is not null)` would equate to "there is someone dead to raise".\
Or\
`if (SimpleTarget.Stack.AllyToRaise is not null)` ("there is someone dead that I care about to raise"), if you're wanting to follow the player's Raise Stack, which is likely wise.

edewen (approved):
Tested on pld. variant actions happen. doesnt seem to break that i can tell

zbee (approved):
Thanks for using #772

PR #767, opened 2025-08-20
[BLM] Standalones

Kagekazu:
- Deconflict standalone combo's


zbee:
Uhhh probably below, I suppose

edewen:
I would say even those larger cooldowns should be on the advanced combo simple bc it needs to be as auto friendly as possible. 

and yeah, giving how big of a shield it is, and the cooldown. the raidwide + the hp slider that could eb set to 70% could be a nice ohshitimnotgettingtoppedoffbcthishealerisdpsingbutton

Kagekazu:
So atm its HP% OR a raidwide is casting, should i make it so it checks BOTH before using it? just doublechecking before i go to bed xD

edewen:
up to you. personal preference tbh. you could have it check both and if people dont wanna slider, they could always set it to 100 to disable slider

PR #768, opened 2025-08-20
[Casters] Addles, and standstill timer rdm

edewen:
- Added Raidwide Addle option to RDM, PCT, SMN
- Added raidwide Magick Barrier RDM
- Added Raidwide Tempura Coat/Grassa PCT
- added slider for time stood still for gapcloser for rdm melee
- removed gap closer from simple for rdm. godspeed young ones


WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -7022,7 +7044,7 @@ public enum Preset
 
     #endregion
 
-    // Last Used 17078
+    // Last Used 17080
```
zbee, line 7047: `17082` actually, it seems

zbee (approved):
Looks like some good adds, and thanks for the time-still slider for the non-dps gap closer.

Kagekazu (approved):

PR #772, opened 2025-08-23
[Content] Optimize Occult Crescent, introduce `ContentSpecificActions`

zbee:
- [X] Upgrade Occult Crescent
  - [X] Removed `ShouldUsePhantomActions`\
        (this caused Phantom logic to run twice when a Phantom action was desired)
  - [X] Upgraded `BestPhantomAction` to a `TryGet`-style method, with a `ref` for the action ID to go out, and the return being a `bool` for whether an action was found\
        (without this `ShouldUsePhantomActions` was required, which caused all Phantom logic to run twice)
  - [X] Reorganized all Phantom Jobs' `region`s into their own methods\
        (may as well, it only makes it harder to search for it if not in the file)
  - [X] Reorganized those methods, and the order in which they are called, to match the CCP\
        (couldn't believe this wasn't already the case)
- [X] Add `ContentSpecificActions`
  - [X] This calls `OccultCrescent.TryGetPhantomAction()`, and `out`s the returned action ID, and has a `bool` for if there is one that should be executed
  - [X] This has placeholders for Variant/Bozja/etc (other content-specific logic), so `ContentSpecificActions.TryGet()` can be called a single time in all jobs, and Variant/Bozja/etc will not need added to every jobs' code.

> @Tartarga re:#765 we can move the condensed logic for Variant actions into a format similar to Occult Crescent, then call that logic in `ContentSpecificActions` - then we can fully remove all Variant code from every job, and we don't have to add any more content-specific logic to any job ever again :D \
> I would appreciate if (once this is merged) you re-oriented #765 to use this new way, as the end-goal.


edewen (approved):
Built and tested a couple classes on dummy. new method works fine

Tartarga (approved):
Code looks good to me

PR #773, opened 2025-08-24
[Core] Add Item Usage

zbee:
- [ ] Build out an initial Item-Usage Framework
  - [ ] Can Support Phoenix Downs and Potions (Stat and eventually maybe Healing)
  - [ ] Works vaguely similar to Action Retargeting
    - [ ] Organized by Presets instead of `replacedAction`?
      - [X] Record Presets from the Combo used
  - [ ] Only works on `All.Item` actions
- [ ] Build out an Inventory Framework
  - [X] Find Item IDs based on a set of parameters
    - [X] Stat Potions
    - [X] Phoenix Down
    - [X] Health Potions
    - [X] Mana Potions
  - [X] Load user's inventory
    - [X] Sorted so that it could be found by the same set of parameters by which the items were found
    - [X] Ordered by the amount of actual stat that items give
  - [ ] Check Inventory for a specified Item ID
  - [ ] `ItemReady()`
- [ ] Build out a Global: Phoenix Down Retargeting Feature for testing the Framework

Will come in another PR, probably:
- [ ] Build out a Global: Auto-Selected Pot Feature (mostly for testing)\
      (would choose the correct pot for the job, and allow choosing "highest tier in inventory" or "second highest tier in inventory"; mainly so you don't have to keep them up to date on your bar)
- [ ] Build out Global: Auto-Use Phoenix Down (Retargeted) Feature\
      via `ContentSpecificActions`, to add auto-rezzing to all jobs\
      would need CDF options
- [ ] Build out Pots in Openers functionality
- [ ] Build out `IsPotSuggestedInEncounter` or something\
      keeps tracks of party member deaths, player being weak/DD'd etc, to advise against potting dead runs / dead burst phases.\
      jobs will use this Property when checking whether to pot before burst.

Unrelated things also done here:
- [X] Occult Crescent's `Revive` action now finally Retargets to the Raise Stack
- [ ] Cleaned up `UseActionDetour`


Kagekazu:
(i am dipping my toes in this to see what i can solve )

PR #778, opened 2025-08-25
[Melees + BLM] Cleanup

Kagekazu:
BLM
- Add forgotten Flare standalone combo

SAM
- Fix standalone options

All
- Cleanup so classes follow same order than CCP


zbee (approved):
LGTM, especially after your most recent commit.

PR #782, opened 2025-08-26
[SAM] Standalone and AoE

Kagekazu:
- Fix Yukikazi standalone option to respect the checkboxes
- Refactor ST and AoE basic combo usage


edewen (approved):
Built and functions. But you know the next thing they will ask for is the true north functionality as an option

zbee (approved):
Much simpler core-combo logic, seems to output just fine, LGTM

PR #783, opened 2025-08-26
LastAction Updates

Taurenkey:
- Moved out of ReceiveActionEffectDetour. Resolves issues where some actions could trigger an effect twice and would count towards certain things twice (Noted in PvP with BRD as an example).
- Interrupted actions now checked via Task created based on cast times, if interrupted the task is cancelled.
- ReceiveActionEffectDetour now reserved for tracking status applications and HP/MP updates. (Possibly other things in the future too).


edewen (approved):

zbee (approved):
Esuna stuff looks good and seems to behave.
`LastAction` stuff doesn't bother DRK in either way it was being affected in its opener before the last change, and Retargeting seems unaffected - so: LGTM

PR #788, opened 2025-08-27
[RDM] Vercure solo

edewen:
Adds a Vercure option to ST and AOE advanced. Will SELF heal when below selected health level if

1. Not in a duty
2. Or if in a Duty you are not in a party (Example: Solo story quest duties)
3. Of if in a duty you are in a fate (Example: Occult Crescent)


Tartarga:
Question, would this flow better if on a Dualcast buff? IIRC, Vercure has a slightly longer cast time than basic like Jolt

edewen:
Vercure is the exact same cast time as jolt, the shortcast. my sheet show 1.95.  I thought it best to have it be the initial cast, and still use the instant cast on the higher potency dps spell.

Tartarga (approved):

zbee (approved):
Cures when I expected it to, doesn't otherwise, LGTM

PR #789, opened 2025-08-27
[BLM] fix fire 1 to despair

Kagekazu:
- Fix fire 1 to despair
- Fix aoe dot usage for pre 58.

### DIFF
--- a/WrathCombo/Combos/PvE/BLM/BLM.cs
+++ b/WrathCombo/Combos/PvE/BLM/BLM.cs
@@ -16,6 +16,12 @@ internal partial class BLM : Caster
             if (actionID is not Fire)
                 return actionID;
 
+            if (Variant.CanCure(Preset.BLM_Variant_Cure, BLM_VariantCure))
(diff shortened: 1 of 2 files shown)


WrathCombo/Combos/PvE/BLM/BLM.cs
```diff
@@ -697,7 +700,7 @@ protected override uint Invoke(uint actionID)
                            !LevelChecked(Fire4) &&
                            HasStatusEffect(Buffs.Firestarter)) && !JustUsed(Fire3) => Fire3,
 
-                Fire when BLM_Fire1_Despair && FirePhase && CurMp < 2400 && LevelChecked(Despair) => Despair,
+                Fire when BLM_F1to3 == 0 && BLM_Fire1_Despair && FirePhase && CurMp < 2400 && LevelChecked(Despair) => Despair,
```
edewen, line 703: despair needs to be above fire 3, otherwise it will only show fire 3.

PR #793, opened 2025-08-31
[MCH] HP%

Kagekazu:
- Add HP% threshold options to various AoE options
- Add HP% slider for ST Hypercharge

### DIFF
--- a/WrathCombo/Combos/PvE/MCH/MCH.cs
+++ b/WrathCombo/Combos/PvE/MCH/MCH.cs
@@ -510,7 +510,7 @@ internal partial class MCH : PhysicalRanged
                     // Hypercharge
                     if (IsEnabled(Preset.MCH_AoE_Adv_Hypercharge) &&
                         (Heat >= 50 || HasStatusEffect(Buffs.Hypercharged)) && LevelChecked(Hypercharge) &&
-                        LevelChecked(AutoCrossbow) && 
+                        LevelChecked(AutoCrossbow) &&
                         GetTargetHPPercent() >= MCH_AoE_HyperchargeHPThreshold &&
                         (LevelChecked(BioBlaster) && GetCooldownRemainingTime(BioBlaster) > 10 ||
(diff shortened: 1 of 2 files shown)


edewen:
Honestly thinking this should wait till next week and more is put into it. 

A slider by itself isn't as helpful. As people will want to set it super low or even zero because of bosses. And rightfully so. 

But a slider with a checkbox of apply this threshold to all content or non bosses only lets the feature actually be useful. No one wants to stop using the cds on bosses. But they would probably want to hold on trash at like 30%

Its a larger change and really shouldn't be rushed as it would need some proper testing. Or at least some creative dummy testing swapping logic around bosses with !.

edewen:
True. Aoe probably doesn't need those non boss vs all content sliders. When you are doing a trash pack. the time it screws you, is when it gets down to that last mob or 2 dependent on settings, and it blows your wad on the st cooldowns. Hence why those should really have the radial button choice. 

When you have no slider, it makes sense to use this on all content or only on bosses. Not ideal, it should have a slider because of solo content really. 

image

But when you are using the slider, the logic should then be "Apply this slider to all content to non bosses only."

image

Kagekazu:
> True. Aoe probably doesn't need those non boss vs all content sliders. When you are doing a trash pack. the time it screws you, is when it gets down to that last mob or 2 dependent on settings, and it blows your wad on the st cooldowns. Hence why those should really have the radial button choice and slider.
> 
> When you have no slider, it makes sense to use this on all content or only on bosses. Not ideal, it should have a slider because of solo content really.
> image
> 
> But when you are using the slider, the logic should then be "Apply this slider to all content or non bosses only."
> image

yeah, ill prob adapt it during next week

edewen:
Things in st that could benefit from having slider/radial next week imo
Barrel stabilizer
Wildfire
Hypercharge
Queen
Reassemble
chain saw

I omit fmf and excavator bc those are limit time procs if you dont use em you end up losing em. Basically anything 60 seconds or above is what I would do.

Tartarga (approved):

PR #806, opened 2025-09-08
[Healers] Fix the priority orders, add default priorities to ccp

edewen:
Fix the priority orders, add default priorities to ccp


Kagekazu (approved):
Code looks good, have not tested in game

zbee (approved):
All looks sound, only the UI stuff tested in game

PR #808, opened 2025-09-09
[SAM] kenki rework + various cleanup

Kagekazu:
- Make all Iaijutsu options toggleable seperatly
- Third Eye raidwide option
- Meditate option
- Add resource class for Kenki
- Add Burst management for Kenki


WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -6333,7 +6357,7 @@ public enum Preset
 
     #endregion
 
-    // Last Value ST = 15023
+    // Last Value ST = 15027
```
zbee, line 6360: Just btw, your last three options jump up to `15093` - ofc beyond the labeled `15027` here.
Doesn't matter, and don't know if you would want to reign in those numbers, or just change this value, or just not mind it, but figured it was worth mentioning.

zbee (approved):
Logic looks generally sound, testing on a dummy seems fine, just a couple unimportant notes above

WrathCombo/Combos/PvE/SAM/SAM_Config.cs
```diff
@@ -126,6 +117,12 @@ internal static void Draw(Preset preset)
                             "Kenki Amount", sliderIncrement: SliderIncrements.Fives);
                     break;
 
+                case Preset.SAM_ST_Meditate:
+                    ImGui.SetCursorPosX(48);
```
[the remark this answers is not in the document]
Kagekazu: > You should always do `48f.Scale()` instead of just `48`

Added in

WrathCombo/Combos/CustomComboPreset.cs
Kagekazu: > Just btw, your last three options jump up to `15093` - ofc beyond the labeled `15027` here. Doesn't matter, and don't know if you would want to reign in those numbers, or just change this value, or just not mind it, but figured it was worth mentioning.

I promise it has logic :P

I reserve 90-99 for healing/mits/role skills, like bloodbath, second wind,.. 
So that when i have to add something to a combo, its not in between all that.

Tartarga (approved):
If it's good enough for zbee it's good enough for me

PR #820, opened 2025-09-21
[RDM] Reprise Feature

edewen:
- New feature (Reprise)
  - Out of range (Simple Mode 5 Yalms), will use Reprise to not break the melee combo as long as there is enough white and black mana to finish the combo after.

### DIFF
--- a/WrathCombo/Combos/PvE/RDM/RDM.cs
+++ b/WrathCombo/Combos/PvE/RDM/RDM.cs
@@ -78,6 +78,13 @@ internal partial class RDM : Caster
                     (HasEnoughManaToStart || CanMagickedSwordplay))
                     return EnchantedRiposte;
             }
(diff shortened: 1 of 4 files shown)


Kagekazu (approved):

zbee (approved):

WrathCombo/Combos/PvE/RDM/RDM.cs
```diff
@@ -78,6 +78,13 @@ protected override uint Invoke(uint actionID)
                     (HasEnoughManaToStart || CanMagickedSwordplay))
                     return EnchantedRiposte;
             }
+            
+            if (LevelChecked(Reprise) &&
```
zbee, line 82: Should this level check actually be for `EnchantedReprise`?

PR #824, opened 2025-09-24
[SAM] More Tweaks

Kagekazu:
- Add better usage for `Senei`, `Zanshin` and `Shoha`
- Add leeway on when to use `Ikishoten`
- Add Substitution for `Gyoten` to `Shinten` when u have enough Kenki in the opener
- Add Dumb Meikyo - use whenever u got 3 sens, no logic
- Add Ogi and Iaijutsu option to ranged

- Move most logic to helper class

### DIFF
--- a/WrathCombo/Combos/PvE/SAM/SAM.cs
+++ b/WrathCombo/Combos/PvE/SAM/SAM.cs
@@ -74,14 +74,14 @@ internal partial class SAM : Melee
             if (UseRanged(true, true))
                 return Enpi;
 
-            if (UseTsubame)
+            if (UseTsubame())
                 return OriginalHook(TsubameGaeshi);
 
-            //Ogi Namikiri Features
+            //Ogi Namikiri feature
             if (!IsMoving() && UseOgi())
                 return OriginalHook(OgiNamikiri);
 
-            // Iaijutsu Features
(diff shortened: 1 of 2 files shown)


WrathCombo/Combos/PvE/SAM/SAM_Config.cs
```diff
@@ -37,22 +37,34 @@ internal static void Draw(Preset preset)
                     break;
 
                 case Preset.SAM_ST_CDs_MeikyoShisui:
-                    DrawHorizontalRadioButton(SAM_ST_MeikyoBossOption,
-                        "Use The Balance Logic in all content", $"Uses {MeikyoShisui.ActionName()} logic regardless of content.", 0);
+                    DrawHorizontalRadioButton(SAM_ST_MeikyoLogic,
+                        "Use No logic", $"Uses {MeikyoShisui.ActionName()} when u have 3 sens.", 0);
```
zbee, line 41: I don't think this is really "No logic".
Maybe "Simple **L**ogic" would be better here

zbee (approved):
LGTM, harder to check the logic is the same/similar when moving to another file though of course.

edewen (approved):

PR #825, opened 2025-09-26
[PCT] Add lvl 90 openers

Kagekazu:
as requested in Discord

i think they are added in correctly

image

image


zbee (approved):
Looks fine. Probably should have handled it with a single `SkipSteps` instead of another Opener, for what its worth.

PR #827, opened 2025-09-28
[Various]  Little Stuff

edewen:
- Added debuff check for OC Cannoneer Silver Cannon
- Fixed lvl 90 pct openers


Kagekazu (approved):
LGTM
I would say for the PCT opener, lock it to 90, im locking the SAM ones aswell.
The only place they are actually useful is locked to that lvl anyway.

PR #833, opened 2025-10-03
[MNK] Unifying stuffs + number tweaks

Kagekazu:
mostly cleanup and making little tweaks so it aligns better for burst windows


edewen (approved):

zbee (approved):
LGTM
Only one `flags` param away from combining the AoE and ST methods into one. Would encourage you check that out in DRK, then you could consolidate the logic even more.

Kagekazu:
> LGTM Only one `flags` param away from combining the AoE and ST methods into one. Would encourage you check that out in DRK, then you could consolidate the logic even more.

yeah, thats probably next step for MNK :P its like my little test subject while i main it this tier :P

PR #848, opened 2025-10-15
[MNK] Cleanup ST and AoE

Kagekazu:
- Merges all the ST and AoE functions and adds the use of parameters
- Add HP% Treshold for Aoe Perfect Balance usage
- Add boss check + hp slider to ST and AoE Buffs instead of seperate
- Removes replies from buff check
- Add better weave check for BH - RoF

### DIFF
--- a/WrathCombo/Combos/PvE/BLM/BLM.cs
+++ b/WrathCombo/Combos/PvE/BLM/BLM.cs
@@ -57,7 +57,7 @@ internal partial class BLM : Caster
 
                 if (IcePhase)
                 {
-                    if (CurMp is MP.MaxMP && JustUsed(Paradox) &&
+                    if (MP.IsFull && JustUsed(Paradox) &&
                         ActionReady(Transpose))
                         return Transpose;
 
@@ -129,7 +129,7 @@ internal partial class BLM : Caster
                     return Fire3;
 
                 if (ActiveParadox &&
-                    CurMp > 1600 &&
+                    MP.Cur > 1600 &&
                     (AstralFireStacks < 3 ||
                      JustUsed(FlareStar, 5) ||
                      !LevelChecked(FlareStar) && ActionReady(Despair)))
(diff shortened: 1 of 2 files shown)


zbee (approved):

WrathCombo/Combos/PvE/MNK/MNK_Config.cs
```diff
@@ -24,28 +25,23 @@ internal static void Draw(Preset preset)
...
-                case Preset.MNK_STUseBrotherhood:
-                    DrawHorizontalRadioButton(MNK_ST_BrotherhoodBossOption,
-                        "All content", $"Uses {Brotherhood.ActionName()} regardless of content.", 0);
+                case Preset.MNK_STUseBuffs:
 
-                    DrawHorizontalRadioButton(MNK_ST_BrotherhoodBossOption,
-                        "Boss encounters Only", $"Only uses {Brotherhood.ActionName()} when in Boss encounters.", 1);
-                    break;
+                    DrawSliderInt(0, 50, MNK_ST_BuffsHPThreshold,
+                        "Stop using at Enemy HP %. Set to Zero to disable this check.");
```
zbee: I didn't realize exactly what this was until I started working on notes, but now that I do: I don't generally agree with this. Maybe it makes sense for MNK with all the Riddles, but I just wanted to put it out there before more sliders were removed elsewhere.

Kagekazu: > I didn't realize exactly what this was until I started working on notes, but now that I do: I don't generally agree with this. Maybe it makes sense for MNK with all the Riddles, but I just wanted to put it out there before more sliders were removed elsewhere.

i did it like this because u never use BH or RoF without eachother, and RoW is used on cooldown whenever available, so in general if ur holding one buff, u wanna hold the rest aswell.

i can make em seperate, but codewise RoF is never used without BH

PR #854, opened 2025-10-19
Consider striking dummies bosses + boss check optimization

Limiana:
I think usually people would want to test their rotations on dummies as on bosses.
Also slightly optimized IsBoss check since it seems like it runs rather often


zbee (approved):
Looks good to me, ~~though I'll change it in #844 to rebuild `BossesBaseIds` on instance change, since a-raids are whiny~~ (forgot that my `Rank` guess was wrong, and it was actually nameplates that were the issue)

PR #877, opened 2025-11-03
[Status] Add invulnerability check M8S adds

Kagekazu:
makes it so multidot doesnt try to dot the invulnerable add


zbee (approved):
Very nice addition

PR #882, opened 2025-11-08
[NIN] More Tuning

Taurenkey:
> [!WARNING]
> BTW this supercedes #881 so that is not required at all.

TCJ methods refactored to remove OriginalHooks to help with potential latency issues.
Simpler Mudras also refactored for similar reasons (the original Simple Mudras still untouched, maybe next week).


edewen (approved):
Simpler mudras seem to work in game. Tested at 100 and 70

PR #883, opened 2025-11-09
[MCH] overcap protection Gauss / Rico

Kagekazu:
- Add overcap protection for all levels for Gauss and Ricochet


zbee (approved):
overcap does make good sense

PR #888, opened 2025-11-16
[BLM] Make 'Scathe' a movement option instead of an individual combo

Dinokin:


Kagekazu (approved):
Then i am fine with it. 
Seems to behave in game aswell.

PR #896, opened 2025-11-18
[Retargeting] Remove `dontCull`

zbee:
- [X] Remove `dontCull`\
      This minor optimization has not in any way been worth the difficulties and bugs it has caused.\
      If the optimization is desired: Using a Target Resolver is infinitely more reliable and optimized than using `dontCull` ever was anyway.
- [X] Remove all usages of it
- [X] Minor formatting in `ActionRetargeting`

---

**Team:**
Your reviews were requested because I removed all usages of `dontCull` from your classes.

**Testing:**
Theoretically this should be possible to just send, zero issue.
But it should probably be tested whether:
- this leads to a notable fps drop (particularly with many Retarget Features enabled, on a healer)
- any of the Retarget Features stop working
- All/Role Features stop overwriting job-specific Features (? shouldn't be anything ? maybe rezzes)

There shouldn't be any other impact, even in the worst case - this parameter didn't affect if Retargets actually _functioned_.

### DIFF
--- a/WrathCombo/Combos/PvE/DRK/DRK.cs
+++ b/WrathCombo/Combos/PvE/DRK/DRK.cs
@@ -379,7 +379,7 @@ internal partial class DRK : Tank
 
             if (target is not null &&
                 CanApplyStatus(target, Buffs.BlackestNightShield))
-                return actionID.Retarget(target, dontCull: true);
+                return actionID.Retarget(target);
 
             return actionID;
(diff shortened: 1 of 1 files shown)


edewen (approved):
Built. Gave it a test dungeon. manual healing feels wrong after so long doing work on auto. but was retargetting fine without cull.

PR #897, opened 2025-11-19
[PLD] Add retarget feature to Cover

Dinokin:


WrathCombo/Combos/PvE/PLD/PLD.cs
```diff
@@ -924,4 +924,26 @@ protected override uint Invoke(uint actionID)
...
+        {
+            if (actionID != Cover)
+                return actionID;
+
+            var target =
+                //Mouseover
+                SimpleTarget.UIMouseOverTarget.IfNotThePlayer().IfInParty() ??
+
+                //Field mouseover
+                SimpleTarget.ModelMouseOverTarget.IfNotThePlayer().IfInParty();
```
edewen: You have 2 directions here. 

1 Make Field Mouseover an option. BLM does this with Aetherial Manipulation, line 623 of blm. Not everyoone likes field MO

2. Leave it non optional field mouseover. In which can you could Simplify things using the mouseover stack like Sage Icarus. line 929
SimpleTarget.Stack.MouseOver Accounts for Both UI and then field.

edewen (approved):

PR #901, opened 2025-11-21
Leviathan (Hard / Extreme) Invincible Checks

Taurenkey:
Free trial so not gonna touch the extreme, but if anyone fancies it's probably just copy/paste with different numbers.


Kagekazu:
I'll unsych the ex later and push em to the PR

PR #911, opened 2025-11-27
[MNK] Add lvl 90 openers

Kagekazu:
- Add lvl 90 versions of LL and SL openers


zbee (approved):
LGTM.
Though, hard to tell if `5` is still the right comparison for `Chakra` or if `2` is the right number of `Perfect Balance` charges for level 90 from just looking at the job guide online.

PR #918, opened 2025-12-03
[RPR] Add Arcane Crest

Kagekazu:
- [x] Add Arcane Crest raidwide option

### DIFF
--- a/WrathCombo/Combos/PvE/RPR/RPR.cs
+++ b/WrathCombo/Combos/PvE/RPR/RPR.cs
@@ -71,7 +71,7 @@ internal partial class RPR : Melee
                 }
 
                 //Auto Arcane Crest
-                if (UseArcaneCrest)
+                if (CanUseArcaneCrest)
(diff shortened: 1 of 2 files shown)


edewen:
its location is problematic. It is a less important mitigation tool and really should not be placed at a highest priority ogcd. Currently at 291, should probably be down lower at 350s with feint. 

Being such a short cd and useful in self healing, the not in party part at least might be worth considering for simple mode as well.

Kagekazu:
> its location is problematic. It is a less important mitigation tool and really should not be placed at a highest priority ogcd. Currently at 291, should probably be down lower at 350s with feint.
> 
> Being such a short cd and useful in self healing, the not in party part at least might be worth considering for simple mode as well.

Ah yeah, i copied it over from SAM, since its in the top spot there aswell.

I can turn it lower and add it to Simple Mode aswell.
Can probably add Tengetsu to Simple Sam aswell

edewen (approved):

PR #920, opened 2025-12-03
[RDM] Update Vercures to Work as per Description

Taurenkey:
Features #13026 and #13222 now work as described.


edewen (commented):
built and tested. Swapping some ! and roles to make doubley sure it works. 

everything looks good. rider does say using ECommons.DalamudServices isnt needed to be  tho.

PR #923, opened 2025-12-05
[Healers] Stuff

edewen:
- [x] Ast: set up the Button selection to match WHM and SCH. Main spam, Dot, Rank 2 of main spam. 
- [x] Sge: Move the dosis 2 selection from the bool to a horizontal radio to match the other 3 healers. (people tend to check all boxes without reading. this should help mitigate peoples idiocy having to choose between one or the other)
- [x] Sge. Zoe Pnuema option has been removed. Learned in whm trying to get misery to happen in the healing combo, it simply does not work to put a dps skill in healer combo. Pnuema acts as a dps needing an enemy target. As AR is now, it is not possible. 
- [x] ST Dots system. All 4 classes
   - [x] Move all 4 classes to the same dot checker function with a compute hp threshold function
   - [x] All 4 healer dots now have 3 sliders. boss, boss adds, trash
   - [x] Sage, removed consideration of aoe dot from the st dot system. a half power dot should never prevent a full power dot from being applied.


edewen:
yeah ill be piping it over to bard at some point too

PR #931, opened 2025-12-12
Opener tweaks

Taurenkey:
- Resets openers properly on job changing in case stuff lingers
- Opener code will wait until the cache has been run through

This only tightens up some things and doesn't look to address any particular problem at hand.


Kagekazu (approved):
Tested in game and works perfectly

zbee (approved):

PR #941, opened 2025-12-15
Some extra IPC utilities

Limiana:
I still need to test event properly. Maybe this PR would be good for testing version


zbee (approved):
Looks good on the stuff I am used to seeing ha.
If you want it in a testing version (which I agree could be good for this `CallGateProvider` thing), then maybe we could do a testing version with this and @Taurenkey's AutoRot upgrades instead of putting them on main right away.

PR #964, opened 2025-12-24
[Debug] Add Content information

zbee:
- [X] Add to debug file
- [X] Add to debug view
  - [X] ~~Add extra here to try to be able to identify why an area is not identified?~~


edewen:
image

edewen (approved):

PR #965, opened 2025-12-24
[RPR] Little Fixes

### LINKED ISSUE #963: Reaper Dynamic True North issue
if Hold True North for Gluttony Option is ticked, the Dynamic True North Feature will hold both charges not just one, effectively stops the feature from working.

Kagekazu:
- Fix Arcance crest going off when not needed
- fix TN not respecting Gluttony option
- Closes #963


zbee (approved):
LGTM, the fix for the linked issue makes sense.

PR #966, opened 2025-12-24
[SAM] Rewrite Meikyo + Higanbana + less strict when to burst

Kagekazu:
- [x] Rewrite Meikyo so its more flexible for downtime. Now it checks senei cooldown instead of a set 6! min rotation (This takes away the boss check/simple meikyo mode)
- [x] Rewrite Higanbana checks so its more freely used in burst, but not lose uptime.
- [x] make burst more fluid, so it optimizes more on what is used when, so it all fits, but doesnt overcap anything.
- [x] Add Skipstep on opener, and extra Gyoten if extra kenki is generated during it.
image


edewen (approved):
code lgtm honestly dont know enough about sam logic though to say if this will appease the needy. not tested in game

PR #967, opened 2025-12-24
[PvP] Action Replacing Off Warning

Tartarga:
Displays warning if action replacing is off for PvP settings.
image


Tartarga:
Made an edit so autorotation would dynamically show

PR #972, opened 2025-12-26
Decouple Autorotation from Hard Targets

Taurenkey:
- [x] Incorporate OptionalTarget from Autorotation into CurrentTarget for maximum compatibility.
- [x] Setting to always hard target has been added and set as a top level config rather than just a damage one.
- [x] Healer retargeting logic adjusted so doesn't cause confusion with checking logic on one thing but not retargeting to said thing in a straight-forward fashion.
- [x] Fix an issue with SCH `Aetherpact` showing up when the Fairy was dismissed.


edewen:
any way to make it hard target ONLY for damage abilities? doomtrains comings and goings make it a bit weird.

Kagekazu (approved):
Haven't had issues while testing in doom ex

PR #974, opened 2025-12-26
[MCH] Reassemble after downtime

Kagekazu:
- makes reassemble for after downtime possible

### DIFF
--- a/WrathCombo/Attributes/OccultCrescentAttribute.cs
+++ b/WrathCombo/Attributes/OccultCrescentAttribute.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using WrathCombo.Combos.PvE;
 
 namespace WrathCombo.Attributes;
 
@@ -6,9 +7,10 @@ namespace WrathCombo.Attributes;
 [AttributeUsage(AttributeTargets.Field)]
 internal class OccultCrescentAttribute : Attribute
 {
-    internal OccultCrescentAttribute(int jobId = -1)
+    internal OccultCrescentAttribute(
+        OccultCrescent.JobIDs job = OccultCrescent.JobIDs.N_A)
     {
-        JobId = jobId;
+        JobId = (int)job;
     }
 
     public int JobId { get; }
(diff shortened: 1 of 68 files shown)


Kagekazu:
> does this go along with #975, or?

Yeah, merge this after, since I'll have to do some cleanup

Taurenkey:
Not sure we want to Reassemble *during* downtime, that can be a waste given how short the buff actually is.

Kagekazu:
> Right now, it looks like it just makes Reassemble have way higher prio than it did before. If it's meant to be used during downtime then check that there's no target around; if it's meant to be used after a downtime, then do something like `!JustUsed(Reassemble, 65)` (to see that it's been sitting ready).

its not meant to be used DURING downtime,
its meant so it doesnt do a tool, and sit on reassemble until a tool is back available..
i only removed the `Not in Combat` check. the prio stays the same..

zbee:
The confusion comes from you explicitly stating "during" downtime in your code.
Yeah, the not in combat check was doing a lot of heavy lifting though. With that check, it was only something that would fire out of combat, naturally, but without that check still there you r are moving Reassemble up from where it currently is to instead be above `Queen`, `Barrel Stabilizer`, `Ricochet`, `Gauss Round`, `Wildfire`, and `Rook Overdrive`, so a lot higher priority.

Kagekazu:
> The confusion comes from you explicitly stating "during" downtime in your code. Yeah, the not in combat check was doing a lot of heavy lifting though. With that check, it was only something that would fire out of combat, naturally, but without that check still there you r are moving Reassemble up from where it currently is to instead be above `Queen`, `Barrel Stabilizer`, `Ricochet`, `Gauss Round`, `Wildfire`, and `Rook Overdrive`, so a lot higher priority.

Added some checks so its only used when not overheated and have weaved nothing.

Also the prio is changing in #975 to be highest prio when not overheated. 
Also `CanReassemble` has checks on its own on when its allowed to use reassemble.

zbee:
Looks like it will at least match #975 now, in essence.

Kagekazu:
> Looks like it will at least match #975 now, in essence.

Yeah, i will prob need to adjust this one once #975 is merged anyway, so ill have a recheck after!

PR #976, opened 2025-12-27
[Status] Tactical Nuke Detection (UAV Inbound)

Tartarga:
Some rudimentary functions. Requires NightmareXIV/ECommons#151 to be merged.


Tartarga:
Going to make an adjustment to the AoE Stack detection later today. Thinking of having it return the target and if it's a multi-hit type similar to the tryget tank buster detection (healer PoV)

Kagekazu (approved):

PR #987, opened 2026-01-01
[Core] Use `ObjectTable` methods more

Tartarga:
Trying to leverage Object Table's functions
Line wrapped AetherPactTarget

### DIFF
--- a/WrathCombo/Extensions/GameObjectExtensions.cs
+++ b/WrathCombo/Extensions/GameObjectExtensions.cs
@@ -1,6 +1,5 @@
 #region
 
-using System.Linq;
 using Dalamud.Game.ClientState.Objects.SubKinds;
 using Dalamud.Game.ClientState.Objects.Types;
 using ECommons.DalamudServices;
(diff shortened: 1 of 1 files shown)


zbee:
Does `.SearchById()` error out or return null if nothing is found?

Tartarga:
and the other, address to object on Svc.Objects

```
    /// <summary>
    /// Create a reference to an FFXIV game object.
    /// </summary>
    /// <param name="address">The address of the object in memory.</param>
    /// <returns><see cref="GameObject"/> object or inheritor containing the requested data.</returns>
    public IGameObject? CreateObjectReference(nint address);
```

zbee:
Now that one definitely can throw

Tartarga:
Found it in Dalamud. SearchById looks cozy safe to me
```
public IGameObject? SearchById(ulong gameObjectId)
    {
        ThreadSafety.AssertMainThread();

        if (gameObjectId is 0)
            return null;

        foreach (var e in this.cachedObjectTable)
        {
            if (e.Update() is { } o && o.GameObjectId == gameObjectId)
                return o;
        }

        return null;
    }
```

PR #989, opened 2026-01-02
[VFX] SharedDamageEffect Cleanup

Tartarga:
All VFX Effect code moved to new file, to separate out of Status which deals with IStatus. New file strictly for VFX Effects

VFX SharedDamageEffect:
- Alliance support for 1 alliance shared (usually multi hit). Old code only worked if it was on a party member
- Detects hidden NPCs having stacks on them (Jade Stoa party member body slam destination)
- Returns a distance based on who/what has the marker (alliance/party/npc)
- The returning PartyMember IBattleChara is optional and does not have an effect on the true/false of the function. Can be used for other things, such as a redirected melee dash to group up, or whatever the team comes up with
- Party members have the highest priority of all these
- Mr. Freeze had a word with the path string lists

Debug: Moved VFX debug checks (TB / Shared) from Party to new VFX area


Kagekazu (approved):
build fine, havent found any errors

PR #1003, opened 2026-01-06
[Various] Cleanup and other stuffs

### LINKED ISSUE #1002: [MNK] Forbidden chakra is never used and requires manual input (both simple and advanced]
This has been happening for awhile but I thought it may be something with my settings. Today I played with the settings and tried both autorotation ON and OFF on both simple and advanced. Set the rotation toggle to bosses and all content (I am mentioning this because the debug file will show one setting, but I tried all) 

It doesn't use it at all.  This is true for striking dummy as well as in content. Tested it with duty support. Trash mobs and boss. 

Does not use it in opener, 2 min burst, nor during filler. 

Wanted to bring it to your attention. Thanks 

debug file below

[WrathDebug.txt](https://github.com/user-attachments/files/24440468/WrathDebug.txt)

### LINKED ISSUE #1005: [MCH] Question about Auto Crossbow
Has Auto Crossbow been removed from the advanced AOE option when overheated is active for a reason? It seems like a wasted button to make it its own button. Been meaning to ask this for a while, just haven't gotten the free moment to do so til now.

Kagekazu:
- [x] General cleanup of my classes, so it uses more helpers on stuff that is both in simple and advanced
- [x] Use chakra skill instead of Originalhooking meditation (for the weirdos that turn off the action changing)
- [x] Add Gapcloser cooldown check in SAM Opener + Add option to skip it completly
- [x] Revamped `Shadow Of Death` on RPR so it is better manageable
- [x] Add Gauss / rico enabled check for Crossbow usage
- [x] Closes #1005 
- [x] Closes #1002
- [x] Add forgotten pool check on non weave reassemble
- [x] Adjust AoE piercing talon option to check for AoE skill range
- [x] Add Manual TN charge for DRG, MNK, RPR and SAM

### DIFF
--- a/WrathCombo/Combos/PvE/RPR/RPR_Config.cs
+++ b/WrathCombo/Combos/PvE/RPR/RPR_Config.cs
@@ -58,7 +58,7 @@ internal partial class RPR
 
                     DrawAdditionalBoolChoice(RPR_ST_TrueNorthDynamic_HoldCharge,
                         "Hold True North for Gluttony Option", "Will hold the last charge of True North for use with Gluttony, even when out of position for Gibbet/Gallows.\n" +
-                                                               "If Above Slider is set, it will NOT use the remaining charge for Gluttony, but for manual use.");
+                                                               "If Above Slider is set to 1, it will NOT use the remaining charge for Gluttony, but for manual use.");
                     break;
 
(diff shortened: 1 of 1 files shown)


edewen (approved):
lgtm not tested in game

PR #1004, opened 2026-01-06
Fix standard step being held

Limiana:
I think I decently tested it now


Limiana:
okay I think this should be good SD logic
but testing is required

zbee:
Formatted it to my liking, but yeah it does look good to me too. Testing always good, idk that it's particularly necessary when you tested the rest.

zbee (approved):
LGTM, even if 1/4 untested.

PR #1008, opened 2026-01-10
Removal of Performance Mode

Taurenkey:
I don't think it serves a purpose anymore. If somehow people are getting performance issues with all the throttlers we have in place then it's probably time to throw out the potato or turn off about 300 other mods and plugins.


Taurenkey:
> [Taurenkey] Performance Mode was very useful for hiding the combos performed by the plugin. It would be good to keep the ability to hide the combos performed by the plugin without the pseudo performance features. If it's possible

Useful for hiding *from what*. Because if you're hiding it from other people then you shouldn't be showing them anything. Even worse if you're using it to trick people. The way it is currently is the way the plugin has always worked, and that's how it is going to stay.

PR #1015, opened 2026-01-13
Queue Overwrite Setting

Taurenkey:
Allows actions already in the queue to be overwritten for more reactive gameplay when combined with bigger queue windows.


Kagekazu (approved):
Tested on VPR and GNB without issues.

PR #1049, opened 2026-02-01
[Auto-Rotation] Just some more adjustments innit

Taurenkey:
Updates the pre-emptive HoT/Shield and cleanse parts of Autorot to use retargeting for better compatibility with standalone retargets (this would take priority and not the heal stack).

Left some comments regarding Kardia & rezzing since they have rules that makes it somewhat trickier to wire up to retargeting without causing more conflicts.

OptionalTarget removed from individual presets as it was causing havoc with autorot and retargeting settings and conflicting. Now it's just one autorot heal target which is referenced and only cleared once actions are fired off.

Moved all one-button healer logic to its own SimpleTarget since retargeting may be disabled and therefore shouldn't use the stack.

Openers will no longer run if the player is occupied, preventing weird stuff with message spam.

### DIFF
--- a/WrathCombo/Combos/PvE/ALL/HealRetargeting.cs
+++ b/WrathCombo/Combos/PvE/ALL/HealRetargeting.cs
@@ -43,64 +43,52 @@ public static class HealRetargeting
     /// Just a shorter reference to
     /// <see cref="SimpleTarget.Stack.AllyToHeal" />
     /// .
-    private static IGameObject? HealStack => SimpleTarget.Stack.AllyToHeal;
+    private static IGameObject? HealStack => SimpleTarget.Stack.OneButtonHealLogic;
 
     /// <summary>
     ///     Retargets the action if the
     ///     <see cref="Configuration.RetargetHealingActionsToStack">
     ///         option to do so
     ///     </see>
-    ///     is enabled.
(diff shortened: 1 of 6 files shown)


zbee (changes requested):
> Moved all one-button healer logic to its own SimpleTarget since retargeting may be disabled and therefore shouldn't use the stack.

This was the point of `.RetargetIfEnabled` before you removed it there.

This approach breaks all other retargeting solutions for healers. A significant loss for literally no reason, especially since this was already fixed until you unfixed it in 4ca608e.
The correct addressing of the current issue is to undo what was done there.

---

If you're intent to break this functionality for literally no reason whatsoever, real or imagined, you have a lot of other content to remove before such nonsense should be merged:
- Any and all language in the settings tab or Retargeting tooltips suggesting that other retargeting plugins can be used.
- Conflicts around other retargeting plugins should be updated to just be flat conflicts if any healing class' actions are enabled, or the current job is a healer.
- The Default+ Healstack options should be removed, only allowing for a custom stack, since there is zero reason for the backwards compatibility.
  - With removing these settings, the Heal Stack returning method, display methods, and settings UI section can all be significantly simplified as well.

---

This is **not** ready for this week and should **not** be approved or otherwise merged.

PR #1055, opened 2026-02-05
Add Opener Status to DTR

Taurenkey:
New option to have the opener status show in the DTR bar for easy access.

(Also introduces Dalamud conflict checks cause y'know, users be usering)


edewen (approved):
Tested in game. Works as advertised

PR #1059, opened 2026-02-09
[PCT] Edewen Presents: Fun with Flags

edewen:
- PCT
  - Major Overhaul to the code. Used flags to have place to edit called by the 4 rotations. 
  - Removed Optimal burst window option in favor of a less rigid natural burst window.
  - Added Option. Holy in white Single target, It will only fire on when hitting 2 targets. Contains charge hold slider.
  - Added Option. Hammer Stamp Movement Adjustment.
    - Slider added to Hammer stamp combo. Set the time remaining on Hammertime buff to finish your combo allowing you to hold it for movement.
    - Adjusted Steel Muse to work the Hammer movement by holding charge until you use it for movement
  - Added option. Burn Boss Slider. Set health of a boss to stop pooling charges and spent what you have left. 
- SGE
  - Quick fix to automatic raidwides. Should have used a level check where I used an actionready for Eukrasia.
- NIN
  - Little ninja tweak to Raiton to hopefully help Avir's non replicatable bug. Not worth a note


edewen:
It has been changed to use the st vs aoe flags and simple vs adv flags so now it can be a combination of flags.

PR #1060, opened 2026-02-10
[RPR][VPR][BLM] Stuffs

Kagekazu:
- RPR
  - Add buff check and moving check for Harpe and Enhanced Harpe.
  - Add suboption to only use Enhanced Harpe.
  - Update range checks for RPR.
- VPR
  - Add check so vicewinder TN only checks the charges to keep when u got normal TN enabled.
- BLM
  - Update triplecast and swiftcase usage outside of movement to not use when about to use manafont.
  - Rework Fire 3 usage to be up to current standard.
  - Rework Paradox to be up to current standard.
  - Add Fire 3 movement option.


edewen (approved):
lgtm. not tested in game

PR #1061, opened 2026-02-11
[IPC] Implement PingPlugin IPC

Taurenkey:
Adds IPC for PingPlugin so we can add it to the debug just for further information to know if latency could be a factor in issues.

Would prefer to do a non-IPC solution for getting the ping but that's out of the scope of this plugin, maybe someone can do it for ECommons if they get bored?


zbee (approved):
Works just fine for me, and lgtm.

PR #1063, opened 2026-02-13
[Core] Add TTK

zbee:
- [ ] Adds Usable TTK Calculations
  - [X] Records DMG changes every 100ms for every target
  - [X] Calculates the TTK using whole-fight DPS (minimum-viable feature)
  - [ ] Returns an improved TTK weighting more recent data more highly\
        (and hopefully trying to reduce the impact of LBs and big dmg change mechs)
  - [X] Adds `GameObject` Extension to access TTK easily
- [ ] Add some TTK Usage as examples
  - [X] Mark enemies invincible if no dmg is sustained in 1.5s
    - [ ] Ask the user to report (or report on their behalf?) when this is detected\
          (to prevent it from becoming the default, which is not nearly as good as actually adding the invincible mechs themselves [instant detection])
  - [ ] Add some example "Dump X if target is nearly dead" Features to jobs
  - [ ] Swap some "Stop using if target HP is <X%" features to TTK ones

> TTK is or was Tauren's project, only making the PR because I pushed it along a bit, and added the invincibility feature.\
> I don't plan to finish the rest of the feature necessarily. So it'll be drafted and an incomplete todo list until Tauren modifies those things, assumedly.


edewen:
Definitely gonna be adding this to picto when it is ready. The motif drawing limit just doesnt work well with health %

Kagekazu:
Same for some MCH queen and SAM shenanigans. Looking forward!

PR #1065, opened 2026-02-15
[SMN] Fun with Flags Part 2

edewen:
- Auto Rotation
  - Changed Max distance dps float to 30 per instructions
- SMN
  - New Option. Added Simple Mode Gapcloser option on AoE. Already existed on ST.
  - No actual user side changes. Converts SMN to the flag system from PCT last week. Working on one job per week from my roster getting the uniform in their ability to be edited in a single rotation. 
- Occult Cresent
  - Removed the 8 unused options from the ccp for OC classes


Kagekazu (approved):
LGTM
have not tested in game

PR #1069, opened 2026-02-17
Fix interrupt delay

Taurenkey:
As it says on the tin


Tartarga:
Out of curiosity why remove the line from the debug?

PR #1070, opened 2026-02-20
Translation of Actions in Presets

Taurenkey:
Presets can now be updated to translate any actions in their titles and descriptions when surrounded by square brackets. Helps with a bit of consistency. e.g Cure -> [Cure], which will then translate to ケアル in JP, Soin in FR and Vita in DE.


Taurenkey:
Putting on hold to address performance concern.

PR #1071, opened 2026-02-20
[OC] Fix Gladiator spamming Defend out of fight

NicolasJOnGithub:
Defend was spammed on CD even if you were AFK.
-> Force the InCombat flag to use it.

The buff is preserved as long as you don't get hit by an auto, which should not happen if you are out of combat.


zbee:
Whatever works, and this is merged; should be fine as is. Thank you for your contribution :)

PR #1073, opened 2026-02-21
Localization Concept

Tartarga:
- All Presets names and descriptions moved to localization resource files. Preset Translations can be started for other languages. Anything not translated defaults to English.
- If game is not recognized as JA, EN, DE, FR, Wrath will try to figure out based on OS (unknown if CN or KR forks add an extra enum (enum=4) for this purpose).
- One example of settings (search box) added as separate resource files. File schema / Direction of the UI TBD
- UserOpts TBD

To Edit in VS, Open CustomComboPresets.resx to see a small spreadsheet-like entry. Entries are enumname_Name & enumname_Desc. During the conversion process, all \n (line breaks) were converted to actual lines breaks. You can do line breaks with Shift+Enter, or you can use \n if you really want.

Note: I have not reviewed all classes since of conversion. I don't see any inline GetActionName / .ActionName like string functions, but those would need to be removed and replaced with English

Draft for now, but appears to be fully functional, tested by loading up Japanese, and going to Roles and Content -> Tank and seeing meme text for "Global Tank Features"

### DIFF
--- a/WrathCombo/Attributes/CustomComboInfoAttribute.cs
+++ b/WrathCombo/Attributes/CustomComboInfoAttribute.cs
@@ -2,7 +2,6 @@ using ECommons.ExcelServices;
 using System;
 using System.Runtime.CompilerServices;
 using WrathCombo.Extensions;
-using static WrathCombo.Window.Text;
 
 namespace WrathCombo.Attributes;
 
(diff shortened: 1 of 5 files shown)


Tartarga:
closed for master pr

PR #1074, opened 2026-02-21
[Readme] Update links for AutoDuty and Questionable

NicolasJOnGithub:
Update links for AutoDuty and Questionable.

Not really sure about what you want for AutoDuty as the previous is archived, but Liza's link has been deactivated. :(


zbee (approved):
Thank you. I should probably update the IPC links too ....

PR #1077, opened 2026-02-22
[Core] Combo Preset Unification / Optimization

Tartarga:
- Setup a single point of reference using Attribute Dictionary, now called AllPresets. Old AllPresets removed
- Frozen Types Implemented
- Attempting to cutdown on reflection, 50k creations of the info attribute vs only having 1807 presets (aiming for 1:1)
- Any additional dictionaries are based upon AllPresets
- Moved from Window/ConfigWindow.cs to Core/Presets.cs under PresetStorage
- Preset sorting groups are now Immutable
- Various calls to Attribute migrated to AllPresets dictionary
- Overriding Dalamud's GetAttribute because reading 1 attribute causes reading/construction of ALL attributes for some reason. ComboInfo Constructor was called up to 50k times, instead of once per preset

### DIFF
--- a/WrathCombo/Core/Presets.cs
+++ b/WrathCombo/Core/Presets.cs
@@ -1,5 +1,4 @@
-﻿using Dalamud.Utility;
-using ECommons;
+﻿using ECommons;
 using ECommons.Logging;
 using System;
 using System.Collections.Frozen;
@@ -9,10 +8,8 @@ using System.Reflection;
 using WrathCombo.Attributes;
 using WrathCombo.Extensions;
 using WrathCombo.Services;
-using WrathCombo.Window.Functions;
 using static WrathCombo.Attributes.PossiblyRetargetedAttribute;
 using static WrathCombo.Core.Configuration;
-using static WrathCombo.Window.Functions.Presets;
 using EZ = ECommons.Throttlers.EzThrottler;
 using TS = System.TimeSpan;
 


WrathCombo/Core/ConfigurationHelper.cs
```diff
@@ -147,7 +147,7 @@ public void ResetFeatures(string config, int[] values)
                         .Where(preset => (int)preset == value)
                         .First();
 
-                    if (!PresetStorage.IsEnabled(preset)) continue;
+                    if (!PresetStorage.IsPresetEnabled(preset)) continue;
```
zbee: Should stay as `IsEnabled`, and this line is a good showcase of why.

WrathCombo/Core/Presets.cs
```diff
@@ -2,29 +2,147 @@
...
 internal static class PresetStorage
 {
-    private static HashSet<Preset>? PvPCombos;
-    private static HashSet<Preset>? VariantCombos;
-    private static HashSet<Preset>? BozjaCombos;
-    private static HashSet<Preset>? OccultCrescentCombos;
-    private static HashSet<Preset>? EurekaCombos;
-    private static Dictionary<Preset, Preset[]>? ConflictingCombos;
-    private static Dictionary<Preset, Preset?>? ParentCombos;  // child: parent
+    private static FrozenSet<Preset>? PvPCombos;
```
zbee: While `Frozen` stuff is better than `Dictionary`s, idk about it being better than a `HashSet`, that seems like a stretch - a `HashSet` should basically be the gold standard, as far as I understand.

WrathCombo/CustomCombo/CustomCombo.cs
```diff
@@ -101,9 +102,9 @@ public unsafe bool TryInvoke(uint actionID, out uint newActionID, IGameObject? t
 
     /// <summary> Invokes the combo. </summary>
     /// <param name="actionID"> Starting action ID. </param>
-    /// 
-    /// 
-    /// 
+    ///
+    ///
+    ///
```
zbee: These should just be deleted.

WrathCombo/Core/Presets.cs
```diff
@@ -43,50 +161,92 @@ public static HashSet<uint> AllRetargetedActions
...
+                var root = attrs.Parent!.Value;
 
-        OccultCrescentCombos = Enum.GetValues<Preset>()
-            .Where(preset => preset.GetAttribute<OccultCrescentAttribute>() != null)
-            .ToHashSet();
+                // Walk up the parent chain until we reach a top-level preset (Parent is null)
+                while (AllPresets[root].Parent.HasValue)
+                {
+                    root = AllPresets[root].Parent!.Value;
+                }
```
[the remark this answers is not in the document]
zbee: Though I will say, Root Parent (or Parent+Grand Parent, like is used in a few files down) should not be the goal for what we check: We should be checking an array of every parent of a Preset. As in, just saving every preset's parent -all the way up to something without a parent- to an array, to be checked against with `.All()`.

WrathCombo/Attributes/CustomComboInfoAttribute.cs
```diff
@@ -27,6 +36,24 @@ internal CustomComboInfoAttribute(string name, string description, Job job, [Cal
...
+        var stack = new StackTrace(1, false).ToString();
+
+        lock (_lock)
+        {
+            if (_seenStacks.Add(stack)) // Add returns true if it was NOT already present
+            {
+                PluginLog.Information(
+                    $"NEW CustomComboInfoAttribute construction site:\n{stack}");
+            }
+        }
```
zbee: Won't need this in final, shouldn't have been pushed.

WrathCombo/Window/Tabs/Debug.cs
```diff
@@ -199,6 +199,9 @@ public enum SheetType
 
         ImGuiEx.Spacing(new Vector2(0f, SpacingMedium));
 
+        CustomStyleText("Enum Count", WrathCombo.ComboCount);
+        CustomStyleText("Info Count", WrathCombo.ComboInfoCount);
```
zbee: Won't need this in final, probably shouldn't have been pushed.

zbee (changes requested):
Still in its earlier stages of course, but this is what I noticed looking at it now.

WrathCombo/Core/ConfigurationHelper.cs
Tartarga: Let me change it back and see. there was an issue somewhere with "ambiguous between that and customcombofunction's version"

WrathCombo/Window/Tabs/Debug.cs
Tartarga: Removed

WrathCombo/Core/Presets.cs
Tartarga: To my knowledge FrozenSet squeaks out ahead on lookups. It might take a split second longer to create, but from what I read it's faster at lookups over long periods of time

PR #1085, opened 2026-03-02
Add CoD Chaotic to Invicibility check

NicolasJOnGithub:
Should fix the "Two target dotting" where it try to apply a DoT to the Add when you are on tiles, or to the boss when you are on the platform.

25/03/2026 : No issue on my side, I correctly DoT the Atomos/Adds when on the platform, and the Boss when on tiles


NicolasJOnGithub:
Thanks! I'll try to go in PF an test it out, once there is a PF up for it. 🥷

PR #1087, opened 2026-03-03
[Content] Consider new Variant Dungeon

zbee:
- [X] Use an ECommons content check instead of checking for specific zones for Variant Dungeons, to match the newest one (and check less often)


edewen (approved):
Tested in game

PR #1115, opened 2026-03-23
[MCH] Add Double mit protection

Kagekazu:
- Add double mit protection, so you dont use both dismantle and tactician in the same weave.


edewen:
This isnt really in the spirit of what was requested. As It could still double mit if the Tactician was used first due to weird cds, or manually. 

The better approach would have dismantle check for self owned tactician. And Tactician check boss debuff for self owned dismantle.

PR #1118, opened 2026-03-24
[NIN PVP] implemented customizable mudra priority system for pvp - both aoe and ST

qnesp1987:
Refactored the Ninja PvP Mudra system to move away from a hardcoded sequence. This change introduces a granular, prioritized Mudra selection system modeled after the existing BLM PvE movement options.

Changes
1- ST Mudra: Replaced the \"Automatic Mudra Mode\" with individual toggles and priority inputs for Huton, Hyosho Ranryu, and Forked Raiju.

2 - AoE Mudra Mode: Added standalone toggles and priority inputs for Huton, Doton, and Goka.

3 - Logic Refactor: Updated the Invoke methods in NINPvP_ST_BurstMode and NINPvP_AoE_BurstMode to dynamically sort and execute Mudras based on user-defined priority while respecting \"Sealed\" status effects and range requirements.

** Meisui was left alone as it was with its own separate slider.

Reason for this change: The previous hardcoded sequence was too rigid. This refactor allows users to customize their Mudra execution order (e.g., prioritizing Forked Raiju over Hyosho) or disabling specific Mudras entirely within the Burst Mode window, providing much-needed flexibility.

### DIFF
--- a/WrathCombo/Combos/PvP/NINPVP.cs
+++ b/WrathCombo/Combos/PvP/NINPVP.cs
@@ -61,7 +61,9 @@ internal static class NINPvP
     {
         public static UserInt
             NINPvP_Meisui_ST = new("NINPvP_Meisui_ST"),
+            NINPvP_ST_FumaShuriken_RangedCharges = new("NINPvP_ST_FumaShuriken_RangedCharges"),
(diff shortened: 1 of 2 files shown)


edewen (approved):
Works in game as expected.

PR #1123, opened 2026-03-28
[Localization] Generics (For April 4th)

Tartarga:
Converting what I can to use Generics translation table


zbee (changes requested):
I think this works within jobs, but when we take them out of jobs like this I believe it becomes only a hurdle (one more thing you have to check, when you're just trying to add a config).
I don't want to see this implemented, left the few other notes I had about this in dev chat, not a 'real' review.

PR #1125, opened 2026-03-28
[WHM]Fix WHM Alternative Raise behavior on Swiftcast cooldown

qnesp1987:
Fixing WHM's alternative raise behavior to work similar to how it does with other healers.

Currently, the button gets stuck on the Swiftcast icon when its on cooldown, and only allows usage with swift which prevents the ability to hard cast the same way AST SGE SCH can.  I've updated the logic to check for the cooldown state so it swaps over to Raise as soon as Swiftcast is used

i assume this looks good. it works for me, but if you see that it should be handed different please reject this. But do know that its an issue with WHM that currently exists using its alt raise feature.

### DIFF
--- a/WrathCombo/Combos/PvE/WHM/WHM.cs
+++ b/WrathCombo/Combos/PvE/WHM/WHM.cs
@@ -711,17 +711,18 @@ internal partial class WHM : Healer
 
         protected override uint Invoke(uint actionID)
         {
-            if (actionID != Role.Swiftcast)
+            if (actionID != RoleActions.Magic.Swiftcast)
                 return actionID;
 
-            var canThinAir = !HasStatusEffect(Buffs.ThinAir) && ActionReady
-                (ThinAir);
+            if (IsOnCooldown(actionID) || HasStatusEffect(RoleActions.Magic.Buffs.Swiftcast))
+            {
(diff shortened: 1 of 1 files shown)


Taurenkey:
Now that I look at these features a bit closer, I think it doesn't actually make much sense to return the raise button whenever Swiftcast is on cooldown since it makes it harder to actually track its cooldown. Maybe I'm missing something, but since this overrides Swiftcast, you would need some sort of external way of tracking Swiftcast cooldown. You still have the rez buttons not being overridden, so idk I can't see why it's like this.

edewen (approved):
Code looks good to me. I would mention thin air has charges so maybe add a !justused thinair just in case, but it didnt have it before and no one was having issues.

qnesp1987:
I just prefer it for less hotbar real estate. however, i agree with @Taurenkey  about not seeing the cooldown. Which is relevant only to those of us who do not one-button everything. as jobs with dps considerations tied to swift often have it in the main rotation. In which this only becomes a "need" when you want to delay your 8 second raise cast and need to know how many seconds you have on Swift - which can easily be solved with a /party macro. i dont know. i personally use the local ones and not the global. 

maybe out of habit since sloth days.

Taurenkey:
Going to merge this as it keeps the functionality similar to the other healers, but I will remove the conflicts between the local ones and the global ones so both can be used at the same time if required.

PR #1127, opened 2026-03-30
[PVP] Couple of movement fixes

edewen:
- RDM
  - Removed movement code at the end locking out jolt while moving
- SMN
  - Removed not moving requirement from slipstream due to changes back in 7.1 its not needed.


Kagekazu (approved):
LGTM

PR #1137, opened 2026-04-12
swap manaward to hp only when solo. + some VPR even minute change

Kagekazu:
- BLM Swap Manaward to just check the HP slider when Solo.
- VPR Make window bigger after `Ouroboros` to weave Reawaken.


Taurenkey (changes requested):
Don't make it HP only, it has infinitely more value when group damage is detected. Just make it OR instead of AND.

Kagekazu:
> Don't make it HP only, it has infinitely more value when group damage is detected. Just make it OR instead of AND.

Yeah, i read the discussion this morning when i woke up. I will adjust this when I'm back home tonight to check raidwide or when solo. (does having companion out also count as solo for our code?)

Taurenkey (changes requested):
So the HP check should really only be applied with the not being in party check, it should always apply on groupwide damage regardless of your current health.

Kagekazu:
> So the HP check should really only be applied with the not being in party check, it should always apply on groupwide damage regardless of your current health.

makes sense :P done!

edewen:
> So the HP check should really only be applied with the not being in party check, it should always apply on groupwide damage regardless of your current health.

Disagree.

Putting an extra shield on yourself when you are already healthy helps no one. If you are already topped with shields, your shield on top of it means you may not need healing after but you are still gonna get it because everyone else will from the aoe healing. 

Running with an hp slider as well, allows it to come in on the situations where you would actually need the shield because you are not topped off and about to get trucked. 

Better it used less often more effectively, than more often just because it exists.

Taurenkey:
> Putting an extra shield on yourself when you are already healthy helps no one.

Key word there is extra. If you have no shield from a healer it will always be effective. The degree of which depends on the strength of the damage, but that's fight dependent so we can't factor that. Manaward is 30% of your max HP, so the value is going to depend on how strong the aoe is. It doesn't matter if you're at 50% or 100% HP, the only value that matters there is the 30% shield and how much of it is going to get consumed. 

@Kagekazu think we can rewrite the feature a bit for advanced. Thinking of keeping the HP slider, but letting users define the behaviours for group play vs solo play differently. I'll probably update the PR at some point this morning with the idea.

edewen:
I get that the shield isnt wasted. But from a healer and math standpoint, all it is doing is making their health % after the hit higher than the rest of the group. They will just end up being overhealed by the following aoe healing and the shield will have really done nothing to help in reality. 

So I agree the hp slider should stay in advanced, even if there are 2 sliders one for group and one for solo. And if a player wants that functionality of every raidwide they can still set it to 100%

Kagekazu:
> > Putting an extra shield on yourself when you are already healthy helps no one.
> 
> Key word there is extra. If you have no shield from a healer it will always be effective. The degree of which depends on the strength of the damage, but that's fight dependent so we can't factor that. Manaward is 30% of your max HP, so the value is going to depend on how strong the aoe is. It doesn't matter if you're at 50% or 100% HP, the only value that matters there is the 30% shield and how much of it is going to get consumed.
> 
> @Kagekazu think we can rewrite the feature a bit for advanced. Thinking of keeping the HP slider, but letting users define the behaviours for group play vs solo play differently. I'll probably update the PR at some point this morning with the idea.

is good! i added something for VPR, since too lazy to make a 2nd PR just for 1 number change.
Lemme know if i have to change something or if u are!

PR #1138, opened 2026-04-12
Fix Mystic Knight spellblades blocking action when out of range

qnesp1987:
Added HasBattleTarget() and InActionRange() checks to BlazingSpellblade, HolySpellblade, and SunderingSpellblade. Without these, the combo button would get stuck trying to execute a spellblade when out of range, blocking all other actions until the player ran into range. Also fixed a whitespace typo on Buffs.BlazingSpellblade.


qnesp1987:
wrong PR. ill recreate.

Issue #1122, opened 2026-03-27
The "Retarget Raise and Vercure" option does not work for Vercure


Wolfahrt:
Hello,

I tried configuring my retarget for Vercure based on my heal stack (UI mousover > soft target > Self) but it does not work at all : vercure is only cast on my hard target.

Note that this seems to work with Raise and it also works with all other healers spells, but not Vercure.

I tried changing with "Roles and content > Magical DPS global option" but it does not work either.

Thanks for help !

Wolfahrt:
Alright, thought Vercure was also counted on the heal stack.
Then sorry this is not an issue but a product enhancement :)
Hope it will be handled soon ! Quite useful for ultimates when boss can't be targeted ^^

edewen:
#1140

edewen:
went live today

PR #1140, opened 2026-04-17
[Tanks] [SMN] [RDM] Edewen Updates

edewen:
- Tanks
   - Added a slider to set a delay on the short mitigation used for each jobs tankbusters. Defaults to zero, which is original functionality. 
-  SMN
   - Added Standalone feature to retarget Rekindle the same way it is done in the main combo. 
- RDM
   - New Standalone Feature: Retarget Vercure with Mouseover and Lowest HPP Party member options. #1122 
-  PLD
   - Added Health Slider check to sheltron overcap. Added generic Minimum gauge to config for me to use with other jobs as well. 
   - Removed HasBattletarget requirement from the bulk of the PLD combos that was causing a break in the cofefe combo when you de-target something, re-engage with fast blade, and then spam holy spirit instead of being able to finish the burst, 
- NIN
   - Removed delayed weave from opener due to bug it was causing with small lags and missing the window locking up opener.


WrathCombo/Combos/CustomComboPreset.cs
```diff
@@ -5514,6 +5514,21 @@ public enum Preset
...
+    RDM_RetargetVercure = 13435,
+
+    [ParentCombo(RDM_RetargetVercure)]
+    [JobInfo(Job.RDM)]
+    [Retargeted(RDM.Vercure)]
+    RDM_RetargetVercure_MO = 13436,
+
+    [ParentCombo(RDM_RetargetVercure)]
+    [JobInfo(Job.RDM)]
+    [Retargeted(RDM.Vercure)]
```
Taurenkey, line 5530: Do these have to be presets rather than configs for the main feature?

edewen: It was copied directly from paladins Clemency retarget setup. They COULD go either way technically. But they still both need to exist in case people wanna retarget to mouseover but not lowest hp, or vise versa.

zbee (approved):
Tanks look good

PR #1142, opened 2026-04-19
[PCT] Edewen Changes

edewen:
- PCT
   - Adjusted Combined Aetherhues Standalone to now affect the Blizzard in Cyan buttons and removed their conflict with the Main combos. Users can now use this as a holding pattern kind of button like other casters do with downranked options.


Kagekazu (approved):
tested in game, looks good

PR #1145, opened 2026-04-23
[Core] Add Burst holding commands

qnesp1987:
Description:
 This PR introduces a QoL improvement for burst management by implementing job-aware /wrath hold and /wrath resume commands.
  
  Rationale:
  Currently, users often refer to the burst holding documentation to create job-specific macros for enabling/disabling
  burst presets. This implementation dynamically maps these actions to the appropriate burst presets for the user's
  current class/job, allowing for a single universal "Hold" and "Resume" macro that works across all roles.
  
  Changes:
   Added BurstPresetMap to handle job-to-preset mapping for combat jobs.
  New command: /wrath hold - Disables all burst-related presets for the current job.
  New command: /wrath resume - Re-enables burst-related presets for the current job.
  
   Benefit:
  Consolidates what used to require dozens of job-specific macros into just two universal commands, significantly
  improving the user experience for multi-job players and reducing macro bloat.

### DIFF
--- a/WrathCombo/Commands.cs
+++ b/WrathCombo/Commands.cs
@@ -840,7 +840,9 @@ public partial class WrathCombo
         var enable = sub switch
         {
             "hold" => false,
+            "disable" => false,
             "resume" => true,
+            "enable" => true,
             _ => !PresetStorage.IsEnabled(presets[0]),
         };
 
--- a/WrathCombo/Commands.cs
+++ b/WrathCombo/Commands.cs
@@ -11,6 +11,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using WrathCombo.API.Enum;
+using WrathCombo.Attributes;
 using WrathCombo.Core;
 using WrathCombo.CustomComboNS;
 using WrathCombo.Data;
@@ -823,6 +824,12 @@ public partial class WrathCombo
     /// </param>
     private void HandleBurstControl(string[] argument)
     {
+        if (!PresetStorage.AllPresets.Any(p => p.Value.JobInfo?.Job == Player.Job && p.Value.ComboType == ComboType.Advanced && PresetStorage.IsEnabled(p.Key)))
+        {
+            DuoLog.Error("This feature is for Advanced Mode Combos.");
+            return;
+        }
+
         if (!BurstPresetMap.TryGetValue(Player.Job, out var presets))
         {
             DuoLog.Error("No burst presets defined for your current job.");
--- a/WrathCombo/Commands.cs
+++ b/WrathCombo/Commands.cs
@@ -50,7 +50,7 @@ public partial class WrathCombo
         { Job.VPR, [Preset.VPR_ST_SerpentsIre, Preset.VPR_ST_Reawaken, Preset.VPR_AoE_SerpentsIre, Preset.VPR_AoE_Reawaken, Preset.VPR_AoE_ReawakenCombo] }, // VPR
         { Job.BRD, [Preset.BRD_Adv_Buffs, Preset.BRD_AoE_Adv_Buffs] }, // BRD
         { Job.MCH, [Preset.MCH_ST_Adv_Stabilizer, Preset.MCH_ST_Adv_WildFire, Preset.MCH_ST_Adv_TurretQueen, Preset.MCH_ST_Adv_Reassemble, Preset.MCH_ST_Adv_Tools, Preset.MCH_AoE_Adv_Reassemble, Preset.MCH_AoE_Adv_Queen, Preset.MCH_AoE_Adv_Stabilizer, Preset.MCH_AoE_Adv_Tools] }, // MCH
-        { Job.DNC, [Preset.DNC_ST_Adv_SS, Preset.DNC_ST_Adv_TS, Preset.DNC_ST_Adv_FanProccs, Preset.DNC_ST_Adv_Feathers, Preset.DNC_ST_Adv_Improvisation, Preset.DNC_AoE_Adv_Devilment, Preset.DNC_AoE_Adv_Flourish, Preset.DNC_AoE_Adv_FanProccs, Preset.DNC_AoE_Adv_Feathers, Preset.DNC_AoE_Adv_Improvisation, Preset.DNC_AoE_Adv_DawnDance] }, // DNC
+        { Job.DNC, [Preset.DNC_ST_Adv_TS, Preset.DNC_ST_Adv_FanProccs, Preset.DNC_ST_Adv_Feathers, Preset.DNC_AoE_Adv_Devilment, Preset.DNC_AoE_Adv_Flourish, Preset.DNC_AoE_Adv_FanProccs, Preset.DNC_AoE_Adv_Feathers, Preset.DNC_AoE_Adv_DawnDance] }, // DNC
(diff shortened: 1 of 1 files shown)


zbee (dismissed):
- [ ] Please refactor to `/wrath burst {hold/resume}`, then `/wrath burst` (with no additional argument) will just toggle it.

- [ ] Please modify the values for `BurstPresetMap` to be `Preset`s directly, e.g. DRK.ShadowBringer, or whatever.

- [ ] Would also appreciate it if the key for `BurstPresetMap` could be of type `ECommons.ExcelServices.Job` instead of just the row ID of the job.
Might make the logic a bit trickier, maybe (should just be able to get `Player.Job` instead of `Svc.ClientState.LocalPlayer.ClassJob`), but would be more consistent with what we've been trying to do.
If it's too much of a headache you can skip this part.

WrathCombo/Commands.cs
```diff
@@ -31,6 +31,31 @@ public partial class WrathCombo
     private const string Command = "/wrath";
     private const string OldCommand = "/scombo";
 
+    private static readonly Dictionary<uint, uint[]> BurstPresetMap = new()
```
zbee: Please modify the values for `BurstPresetMap` to be `Preset`s directly, e.g. `Preset.DRK_ST_CD_Delirium` instead of `5015`, or whatever.

zbee: Please change the key for `BurstPresetMap` to be of type `ECommons.ExcelServices.Job` instead of just the row ID of the job.

Might make the logic a bit trickier (should just be able to get `Player.Job` instead of `Svc.ClientState.LocalPlayer.ClassJob`), but would be more consistent with what we've been trying to do.

WrathCombo/Commands.cs
```diff
@@ -66,6 +93,10 @@ private void OnCommand(string command, string arguments)
         var argumentParts = arguments.ToLowerInvariant().Split();
         switch (argumentParts[0])
         {
+            case "hold":
+            case "resume":
```
zbee: Please refactor to `/wrath burst {hold/resume}`, then `/wrath burst` (with no additional argument) will just toggle it.

WrathCombo/Commands.cs
```diff
@@ -40,6 +65,8 @@ private void RegisterCommands()
         EzCmd.Add(Command, OnCommand,
             "Open a window to edit custom combo settings.\n" +
             $"{Command} auto → Toggle Auto-rotation on/off.\n" +
+            $"{Command} hold → Disables all burst presets for your current job.\n" +
```
zbee: Please remove the command from here, and add it to the Readme commands section instead.
This section will be revisited once a help tab is added.

zbee (changes requested):

qnesp1987:
Addressed all four comments switched the map to `Dictionary<Job, Preset[]>`, moved to `Player.Job`, refactored to `/wrath burst {hold|resume}` with bare `/wrath burst` toggling, and removed the help text. Let me know if anything else needs changing.

Taurenkey:
Food for thought, whilst anyone using this command should already have the initial advanced mode(s) enabled, it might be worth forcing that parent combo on during the enable stage in case it's disabled (either due to simple mode being on or just neglecting to turn it on to begin with).

qnesp1987:
> Food for thought, whilst anyone using this command should already have the initial advanced mode(s) enabled, it might be worth forcing that parent combo on during the enable stage in case it's disabled (either due to simple mode being on or just neglecting to turn it on to begin with).

Good catch. I think the most logical way to address this is to specify its functionality to advanced modes, so if a user tries it without having parent advanced mode active for any of the reasons you mentioned above, it would state in an error message that advanced mode is required for this option to work. I think this is the simplest solution, since the alternative is to either apply it to simple mode's burst that has no toggles, or to incorporate all advanced mode toggles into the function where /wrath burst resume would turn all on and /wrath burst hold would turn all on except for burst options (both disabling simple mode). 


Thoughts? Happy to try either, or any other option you think is better.

zbee:
> it might be worth forcing that parent combo on during the enable stage

This is already done. See `WrathCombo.Core.PresetStorage.EnablePreset()`.

>  so if a user tries it without having parent advanced mode active for any of the reasons you mentioned above, it would state in an error message that advanced mode is required for this option to work.
> 
> Thoughts? Happy to try either, or any other option you think is better.

I'd say do this:
- If Advanced mode **is not** enabled when disabling burst: error out, "This feature is for Advanced Mode Combos"; similar if Simple mode **is** enabled when enabling burst: error out the same. (using DuoLog)
- Keep using `WrathCombo.Core.PresetStorage.EnablePreset()` to enable the presets as you are. It already enables parent presets as required, and disables conflicts.

qnesp1987:
@zbee  Refactored the command to include the Advanced Mode check and the specific error message as discussed. Also switched to PresetStorage.EnablePreset to handle parent dependencies. please let me know if there's anything else.

WrathCombo/Commands.cs
```diff
@@ -786,4 +815,43 @@ private void HandleJobStoneCheckCommand()
...
+
+    /// <summary>
+    ///     Handles the burst control command, toggling, holding, or resuming all burst presets for the current job.
+    /// </summary>
+    /// <param name="argument">
+    ///     The subcommand: <c>hold</c>, <c>resume</c>, or blank to toggle based on current state.
+    /// </param>
+    private void HandleBurstControl(string[] argument)
+    {
+        if (!PresetStorage.AllPresets.Any(p => p.Value.JobInfo?.Job == Player.Job && p.Value.ComboType == ComboType.Advanced && PresetStorage.IsEnabled(p.Key)))
```
zbee, line 827: You'll have to move this further down, where you can check if they are enabling or disabling burst, like I said.

If (`enable` from line 840) `enable && NO_ADVANCED_MODE_ON` OR `!enable && SIMPLE_MODE_ON`.

zbee (changes requested):

zbee (approved):
Thanks for implementing those changes, this looks pretty good to me.

Only other thing I'd do is make a "disable" alias for "hold" and an "enable" alias for "resume", just so they could also be used in the same manner as other commands.

qnesp1987:
added enable and disable as requested. 

also tweaked DNC by removing standard step and improv from burst as i dont think standard step should ever be off (thinking of prolonged holds like bubble phase in m10s) and improv isn't  a burst ability.

zbee:
Put SS back, it's what controls DNC's burst starting. It shouldn't ever be stopped ..... unless you're holding burst.

qnesp1987:
What would you suggest for phases like the bubble phase in m10s where disabling this option will lead to your buff dropping, and to reapply it will mean to either drift your technical step out of raid buffs or to start tech without it? that phase is either longer than the whole duration of the buff or if not, it will certainly drop during your tech. 

my reasoning is to always keep standard dancing on to mitigate such things from happening. This also applies to ultimates where downtime timings can vary and some of which will face the same issue, leading to significant dps loss. 

ill be happy to put it back, but would like to know what you think of this first and if maybe we can find some workaround specific to DNC. 

to clarify, adding it back means DNC burst off would just be cascade filler GCD. Nothing else.

zbee:
It's too early for me, so I don't really have an answer for you. You can open another PR for that commit for me to dig into at another time, but let's not have this hold up this PR 👍🏻

qnesp1987:
done

PR #1151, opened 2026-05-01
[ALL] Roles & Content Icon, Wrong PvP Warning

Tartarga:
Updated the number to use the previous multi color icon
image

Fixed wrong PvP warning being displayed for Action Replacing


Taurenkey:
I kinda liked the one we had, matched the other icon, but alas, tis fine

PR #1154, opened 2026-05-02
[RDM] Combo Adjustments and Localized Configs

Taurenkey:
- Adjust the standalone combos to only return its respective element for Verflare/holy rather than adjust dynamically to match the functionality of the other options.
- Convert config to localized parameters.


WrathCombo/Combos/PvE/RDM/RDM.cs
```diff
@@ -555,11 +555,17 @@ protected override uint Invoke(uint actionID)
...
+                if (ActionReady(Scorch))
+                    return Scorch;
 
-            if (RDM_VerAero_Options[0] && HasManaStacks)
-                return UseHolyFlare(actionID);
+                if (ActionReady(Resolution))
+                    return Resolution;
+            }
+
+            if (RDM_VerAero_Options[0] && ActionReady(Verholy))
```
edewen, line 567: This is not as simple as choosing which mana to build. You get more mana using the correct flare/holy. There is an objectively correct choice so the smart code makes sense to exist.

This needs to be verholy with the option to be smart, and the other one verflare with the option to be smart if we are giving them the choice to manage that on their own.

edewen (changes requested):

WrathCombo/Combos/PvE/RDM/RDM.cs
zbee: Completely changing the purpose of standalone combos seems like it'd do more harm than good, even **if** these might be good changes.
These changes should just be new standalone combos.

edewen: nah its just changing a part of the standalone, but still a part that has existed and has been used. Just make smart an option, that defaults to on, and we good. 

We dont know how many people are using it and happy with it in its current state, but it's safe to assume it is more than 0 so I don't see the logic in pulling the rug from under them to appease one person in general chat when we can make everyone happy with an option.

edewen (approved):
Tested. The option works fine. Ty

PR #1157, opened 2026-05-06
Fix for BossMod IAsyncDalamudPLugin stacktrace

PhoenixtheII:
zbeeRole icon, publisher ♢ — Yesterday at 11:16 PM
now you know phoenix
pls fix

Okay :3


zbee (approved):
Added some more handling for it, that should do nicely for now. Might need more work later (as dalamud stabilizes)

PR #1160, opened 2026-05-06
[PLD] Adjustments on Requiescat Spender and Spirits within / Circle of Scorn Features

Dinokin:
Today I was playing PLD and noticed the Requiescat Spender and Spirits Within / Circle of Scorn features weren't working properly. After some investigation, I observed that those options start at 0 instead of 1, so I adjusted the logic accordingly. I also added Goring Blade to the Requiescat Spender when using the Add Fight or Flight option because I think it makes sense, but I can remove that if necessary.


edewen (approved):
Ty. Not sure how this made it past any basic testing process.

PR #1161, opened 2026-05-07
[Pyretics] Clyteum / Motion Scanner

Tartarga:
New code identifies the hidden NPC on the battlefield and it's progress across it. Disabled Chat based detection, and removed Motion Tracking status from StatusChanged watching.

### DIFF
--- a/WrathCombo/AutoRotation/AutoRotationController.cs
+++ b/WrathCombo/AutoRotation/AutoRotationController.cs
@@ -58,15 +58,13 @@ internal unsafe class AutoRotationController
     public AutoRotationController()
     {
         OnPartyCombatChanged += ResetError;
-        //Svc.Chat.ChatMessage += ScanForWarnings;
+        Svc.Chat.ChatMessage += ScanForWarnings;
         OnStatusChanged += StatusChanged;
     }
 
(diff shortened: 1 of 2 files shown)


Taurenkey (changes requested):
You removed the event for scanning messages (which I'll want to keep for further updates btw) but commented out the check on the removal of the motion tracking buff and kept the event scanning for status changes?

Does your action penalty changes completely handle the mechanic without touching autorot itself? If so, can you please restore the chat message event subscribers and just remove the Clyteum stuff from it.

Tartarga:
> You removed the event for scanning messages (which I'll want to keep for further updates btw) but commented out the check on the removal of the motion tracking buff and kept the event scanning for status changes?

> Does your action penalty changes completely handle the mechanic without touching autorot itself? If so, can you please restore the chat message event subscribers and just remove the Clyteum stuff from it.

The code is under ShouldSkipAutorotation -> PlayerHasActionPenalty with Pyretics and Acceleration Bomb, so when true, AutoRot will just not run at that moment. I commented out your fix to make sure it would work under PlayerHasActionPenalty only.  

I removed the tracker status resume as I knew the status change handler would be useful, but I disabled the chat handler in general as I was unsure. I'll reenable with the dungeon code removed shortly

PR #1164, opened 2026-05-09
[Debug] Add job quests to debugfile

Taurenkey:
Adding quests related to action unlocks into the debug file just to verify it's not something that mundane causing issues.

### DIFF
--- a/WrathCombo/DebugFile.cs
+++ b/WrathCombo/DebugFile.cs
@@ -390,21 +390,21 @@ public static class DebugFile
                 ["Rotation Behavior"] = new Dictionary<object, object>
                 {
                     // Key in Settings           Alias for Setting
-                    ["BlockSpellOnMove"] = "Block Spell on Move",
-                    ["ActionChanging"] = "Action Replacing",
-                    ["SuppressQueuedActions"] = "Queued Action Suppression",
-                    ["Throttle"] = "Throttle (ms)",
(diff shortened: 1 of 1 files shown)


WrathCombo/DebugFile.cs
```diff
@@ -255,6 +260,29 @@ private static void AddPlayerInfo()
...
+    {
+        var player = Player.Object;
+        var job = (Job)player.ClassJob.RowId;
+        var actions = Svc.Data.GetExcelSheet<Lumina.Excel.Sheets.Action>().Where(x => x.IsPlayerAction && x.UnlockLink.RowId != 0 && x.ClassJobCategory.Value.IsJobInCategory(job)).OrderBy(x => x.ClassJobLevel).ToList();
+        AddLine("JOB QUESTS");
+        try
+        {
+            foreach (var a in actions)
+            {
+                if (a.ClassJobLevel <= player.Level)
```
Tartarga, line 273: Why setup the player and job vars when IsJobInCategory(Player.Job) and a.ClassJobLevel <= Player.Level would do?

Tartarga (approved):

WrathCombo/DebugFile.cs
Taurenkey: Job is used to filter to current job in the actions LINQ. Also I just decided to filter out quests the user wouldn't be able to do yet due to level just to help filter out unnecessary noise.

PR #1168, opened 2026-05-10
Simple tweaks

Kagekazu:
- Update VPR opener for 7.5.
- Update RPR `Shadow Of Death` usage for `Double Enshroud` window.
- Update RPR Opener to require 0 gauges, so it doesnt go off mid fight.
- Update MNK `Basic Combo` to use the correct skill during `Perfect Balance` windows.
  - Add `Masterfull Blitz` option to the basic combo.
  - Add `Chakra` usage to basic combo.
- Update `Meditation` usage in AoE Combo to be when no enemies around instead of lower than 3. 
- Add option to MCH to choose to use both `Gauss Round` and `Ricochet` or just `Gauss Round` ( for solo Deep Dungeons)


WrathCombo/Combos/PvE/MCH/MCH_Config.cs
```diff
@@ -124,17 +124,26 @@ internal static void Draw(Preset preset)
...
+                        DrawSliderInt(0, 2, MCH_ST_GaussRicoManualUse,
+                            Generics.ChargePool);
+                    }
+
+                    DrawHorizontalRadioButton(MCH_ST_GaussOnlyOrBoth,
+                        FormatAndCache(Generics.Use0And1, GaussRound.ActionName(), Ricochet.ActionName()),
+                        FormatAndCache(Generics.Use0And1, GaussRound.ActionName(), Ricochet.ActionName()), 0);
+
+                    DrawHorizontalRadioButton(MCH_ST_GaussOnlyOrBoth,
+                        FormatAndCache(MCH_Config.OnlyUse0NotRecommended, GaussRound.ActionName()),
```
edewen: Personally here. I would have "Use only gauss" as the name on the button, and save the wordy part for the description only. This way you aren't adding a small paragraph always in sight on the config. There is not much point in a description on hover if it says the exact same thing as the name of the button.

Kagekazu: Ah ye, I'll swap it when I'm home!

Taurenkey:
Silly Kage, SimpleTweaks is another plugin.

edewen (commented):
lgtm. not tested in game

edewen (approved):

PR #1170, opened 2026-05-11
[Content] Add damage buff only usage for OC abilities

zbee:
- [X] Add global Bursting detection
- [X] Moves DRK's checking for Burst Phase to use new detection
- [X] Adds feature to limit all damaging Phantom abilities to be within a buff of any sort (including smaller stuff, like PLD's FoF)
- [X] Adds warning to other costly OC features
- [X] Removes DRK's bait Darkside status check (it can only be checked through the gauge)

(tested by requesting user)

### DIFF
--- a/WrathCombo/Combos/CustomComboPreset.cs
+++ b/WrathCombo/Combos/CustomComboPreset.cs
@@ -530,6 +530,10 @@ public enum Preset
     #endregion
 
     #region PHANTOM ACTIONS
+    [OccultCrescent]
+    [JobInfo(Job.ADV)]
+    Phantom_RestrictToBuff = 109999,
+
     [OccultCrescent(OccultCrescent.JobIDs.Freelancer)]
     [JobInfo(Job.ADV)]
     Phantom_Freelancer = 110000,
--- a/WrathCombo/Combos/PvE/ALL/Bursting.cs
+++ b/WrathCombo/Combos/PvE/ALL/Bursting.cs
@@ -78,8 +78,8 @@ public class Bursting
                     HasBuff.Target(NIN.Debuffs.TrickAttack) ||
                     HasBuff.Target(NIN.Debuffs.KunaisBane) ||
                     HasBuff.Target(BRD.Buffs.RagingStrikes) ||
-                    HasBuff.Target(BRD.Buffs.RagingStrikes) ||
                     HasBuff.Target(BRD.Buffs.RadiantFinale) ||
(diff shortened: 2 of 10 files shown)


edewen:
This would need a LOT more planning into the OC skills and how they actually function to be worthwhile.

Monk: 30 sec cds
Thief: Pilfer wep doesn't actually have a potency
Sam: Zeninage 2 min cd could actually benefit, but ianuki is a 40 sec cd. 
Berserker: rage works based on taking damage to increase the damage, locking that to a burst window makes no sense
Ranger: Aim is just a 2 min buff so this would work to allign with the buff window. 
Time mage: Comet at 60 seconds might work for a pld with a 60 sec buff, but for most they will throw away half thier comets
Bard: heros rime would work as a 2 min cd, but offensive aria is what you keep up at all other times you dont want that locked
Oracle: Another 60 second cd, but also with the actual predict mechanics, if you dont choose one, you die. so if it takes you outside of a buff window and you havent chose, it wont and you die.
Canoneer: All 30 sec cds. Huge ones at that, most would be skipped. And silver sickness would not be kept on the boss
Mystic KNight: Same boat as canoneer with the 30 second cds and debuff/selfbuff maintenence
Dancer: More 30 sec cooldowns
Gladiator: 2 min cds, would work with it. 

"Many jobs do not have self buffs that are up regularly, many burst phases cannot accommodate extra Phantom Abilities, and many Phantom Abilities are up more often than this on would be on many jobs!"

If this disclaimer even needs to exist, then the feature shouldn't. The only time holding for buffs would even make sense is the abilities that would naturally align with 2 mins due to their own 120 second cooldown. Aim, Heroes Rime, Zeninage, Long reach, and blade blitz. Anything with a shorter cooldown you are just skipping multiple casts to make one marginally stronger. 

I wont request changes. I just won't be the one to approve it given my stance it shouldn't exist.

zbee:
Your mention of all the shorter cooldowns of the phantom abilities completely overlooks that nearly half of base jobs in the game will trip the damage buff check every 30 seconds or more, AND it counts party buffs too, which nobody's aligning in there, and many people are partying for a lot of the content.

I'll add a second warning that clarifies that it will skip casts if you're not on one of those jobs though.

zbee:
The requesting user has very thoroughly tested this, and we found several things to fix or improve.

Kagekazu (approved):
LGTM

PR #1175, opened 2026-05-21
[SGE] Eukrasian Dykrasia Tweak

Taurenkey:
- Updates Eukrasian Dykrasia check on simple AoE to not require a target and just evaluates targets around the player.

### DIFF
--- a/WrathCombo/Combos/PvE/SGE/SGE.cs
+++ b/WrathCombo/Combos/PvE/SGE/SGE.cs
@@ -145,7 +145,9 @@ internal partial class SGE : Healer
                     return Soteria;
             }
 
-            var hasDotTarget = EnemiesInRange(EukrasianDyskrasia).Any(x => GetPossessedStatusRemainingTime(Debuffs.EukrasianDyskrasia, x) is <= 4 or float.NaN && GetTargetHPPercent(x) > 25);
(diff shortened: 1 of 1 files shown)


edewen:
From my understanding this is looking for 4 or more undotted targets? Would prefer 5. 4 is only SLIGHTLY better than ST, but if you are considering overwriting any st dot, which you would be in real game application, 5 is just the safer bet. 

Could go deeper on the math but no reason to super over-complicate the code for minmaxing dungeon trash to make sure it only overwrites the st dot if there is only 1 out of 8 targets dotted but still fires if there is no overwriting. Not worth it imo. 

Also, Aoe Advanced would need this change added as well.

WrathCombo/Combos/PvE/SGE/SGE.cs
```diff
@@ -373,17 +373,16 @@ protected override uint Invoke(uint actionID)
                     return Soteria;
             }
 
+            var hasDotTarget = EnemiesInRange(EukrasianDyskrasia).Any(x => (GetPossessedStatusRemainingTime(Debuffs.EukrasianDyskrasia, x) is <= 4 or float.NaN &&
```
edewen: You mentioned the any can be a count? that would still need to be at least 5 or we will be immediately overwriting the run up dots causing a large potency loss.

Kagekazu (approved):

PR #1177, opened 2026-05-25
[CORE] Optimizations

Kagekazu:
**Action hook (`ActionReplacer`)**
- `FilteredCombos` is materialized once as `CustomCombo[]` when the list is rebuilt (same triggers as live: null on first use, then job/territory refresh) — avoids repeated `ToArray()` on each rebuild
- Disabled-job checks use `HashSet<Job>.Contains` instead of `List.Any`

**Auto-rotation**
- Action selection: single-pass partition into AoE/ST scratch lists (replaces LINQ sort/filter per tick)
- `ProcessAutoActions` / `ProcessPartition` return `true` when an action is queued (fixes inverted return path used by OOC bypass buffs)
- DPS target list built once per tick and shared; pre-emptive HoT/shield use one hostile scan
- Auto-rotation config wrappers (DPS/healer settings) cached on the controller instead of recreated every read

**Per-frame / combo runtime**
- DTR opener text and auto-rot DTR only rebuild `SeString`s when values change
- Party lookups, status/death tracking, timers: reuse scratch storage; fewer list copies
- Targeting: circle/cone/line use the known target where possible; removed unused helper; status cache simplified for common checks
- `CustomCombo` shared constants stored once per type

### Bug fixes
- **Bypass buffs (OOC):** `Run()` now exits after a buff is applied (`ProcessAutoActions` success was always `false` on live)
- **Circle AoE:** origin used wrong reference point
- **Has Evasion Up:** was checking Damage Up buffs

### DIFF
--- a/WrathCombo/AutoRotation/AutoRotationController.cs
+++ b/WrathCombo/AutoRotation/AutoRotationController.cs
@@ -48,7 +48,8 @@ internal unsafe class AutoRotationController
 
     public static bool WouldLikeToGroundTarget;
     public static bool Paused;
-    public static int UnpauseSeconds;
+    // Placeholder for the unimplemented territory-warning auto-pause feature in ScanForWarnings.
+    public static int UnpauseSeconds = 0;
 
     public static IGameObject? AutorotHealTarget;
     public static bool AutorotRaidwiding;
@@ -73,7 +74,6 @@ internal unsafe class AutoRotationController
             return;
 
         bool pauseWarningFound = false;
-        bool raidwideWarningFound = false;
         var logMessages = Svc.Data.Excel.GetSheet<LogMessage>();
         switch (Content.TerritoryID)
         {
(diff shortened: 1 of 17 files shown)


Kagekazu:
have a test/play with it and lemme know if anything needs changing. so far i've had no issues.
It should make ppl have some more frames

Taurenkey:
> FilteredCombos is now an array instead of an unmaterialized IEnumerable — was re-running the Where filter over ~200 combos on every single hook call

This wasn't happening anyway, the first thing the hook does is check if the list is null and updates it accordingly, so it only ran the filter on the very first hook attempt (otherwise it would defeat the purpose of having a filtered list and caching it).

zbee:
With that example, I'd just like to say that this really needs to wait through thorough thorough testing. I've re-subbed to put in some of that work.

Kagekazu:
Oh ye, this is at least a week or 2 to test at least

WrathCombo/AutoRotation/AutoRotationController.cs
```diff
@@ -411,8 +411,8 @@ private static bool ProcessAutoActions(Dictionary<Preset, bool> autoActions, ref
             }
         }
 
-        if (ProcessPartition(_processAoEScratch)) return false;
-        ProcessPartition(_processSTScratch);
+        if (ProcessPartition(_processAoEScratch)) return true;
```
zbee, line 414: This looks like it could be a mistake

Taurenkey:
> Combo list for your current job is built once and reused, not rebuilt every button press

Another confusing thing that isn't happening on current. Combo lists on live are only built on a couple of occasions:

- Before attempting to access the list in the adjusted action hook if it's not already initialised (as per earlier comment)
- On job or territory change as that is where sweeping changes actually happen, job switching is self-explanatory and territory for switching between PvE and PvP.

So essentially you're always walking about with a cached list of all your jobs combos, not just the enabled ones as this gets checked further down the line, so I'm not sure where the idea of it getting rebuilt on every button press comes from as that's not how the adjusted action hook works.

Kagekazu:
> > Combo list for your current job is built once and reused, not rebuilt every button press
> 
> Another confusing thing that isn't happening on current. Combo lists on live are only built on a couple of occasions:
> 
> * Before attempting to access the list in the adjusted action hook if it's not already initialised (as per earlier comment)
> * On job or territory change as that is where sweeping changes actually happen, job switching is self-explanatory and territory for switching between PvE and PvP.
> 
> So essentially you're always walking about with a cached list of all your jobs combos, not just the enabled ones as this gets checked further down the line, so I'm not sure where the idea of it getting rebuilt on every button press comes from as that's not how the adjusted action hook works.

yeah, bad wording on my part xD

PR #1180, opened 2026-06-01
[ConflictingPlugins] Update BossMod IPC

zbee:


zbee:
#1207 completes this work

PR #1181, opened 2026-06-04
Variant api

Kagekazu:
Add Variant skills to the IPC

@zbee do have a check and let me know if it needs any changes!

is together with https://github.com/PunishXIV/WrathCombo.API/pull/4


zbee (approved):
Bumped .API, removed Signature changes in both PRs (since that was ill-advised).
Thanks for showing me your tests and making my privately-requested changes already.

PR #1182, opened 2026-06-04
[MCH] Updates on Reassemble 

Kagekazu:
- Fix `Reassemble` going off on `Hotshot`.
- Rewrote `Reassemble` logic so it always uses it right after eachother in even windows.
- Add option for opener to start without battletarget
- Update AoE `Reassemble` usage


zbee (approved):
LGTM; from following the conversation with Tauren with you on this I assume it's even more LGTM, and with Lim I assume it also tested well.

PR #1183, opened 2026-06-05
Extra mch options

Limiana:
Queen use in Hypercharge
Drift Wildfire over Tools if tools cooldowns are less than this value
Allow using Chainsaw after Wildfire
Allow using Excavator after Wildfire

I'm really not sure what's up with these designer files though


Kagekazu (approved):
LGTM

Kagekazu:
Also fixed the resx layout files

PR #1187, opened 2026-06-08
[WAR][MNK-PvP] Edewen Fixes

### LINKED ISSUE #1186: Primal Rend stuck active in Fell Cleave features and before it is learned
On my level 51 Warrior, the fell cleave features being enabled forces primal rend to the forefront, despite not being active nor learned. It happens even despite the option being turned off, following the same stationary movement check setting as if it was on. This makes the rest of the features unusable when the movement conditions in the setting are met.

edewen:
Fixed primal rend bug with fell cleave, decoded the parenthesis spaghetti. 
Closes #1186 

Fixed Chaotic Cyclone showing Decimate icon. (missing OH) 

Fixed Monk pvp Rising Phoenix not firing in real pvp. Enemiesinrange apparently not working on players but does on dummies.

### DIFF
--- a/WrathCombo/Combos/PvE/WAR/WAR.cs
+++ b/WrathCombo/Combos/PvE/WAR/WAR.cs
@@ -184,29 +184,53 @@ internal partial class WAR
         {
             if (action is not (InnerBeast or FellCleave))
                 return action;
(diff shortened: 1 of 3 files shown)


Taurenkey:
> Enemiesinrange apparently not working on players

I can see why, at some point the check is only looking at BattleNPCs which players obviously aren't. I'll take that bit out, and it's up to you if you want to change anything to reflect it.

PR #1190, opened 2026-06-11
Stop attacking TEA Jagd Dolls below the 25% feed threshold

loop0x539:
Jagd Dolls in The Epic of Alexander (Ultimate) must be brought below 25% HP to be fed to Liquid Rage, but killing one wipes the raid.

- TargetIsInvincible: ignore dolls under 25% so Auto-Rotation targeting, AoE enemy counting, and retargeting skip them
- EncounterSafety.TryGetTEADollBlock: block combo buttons (via Savage Blade replacement) while the hard target is a doll under 25%

This was tested in A4N since it's hard to find a group for TEA right now but can once more players are doing other content than UMAD. I have some screenshots when I was testing it and even used the debug tool in the addon. Feel free to DM me on discord.
Discord: loop_0x539

### DIFF
--- a/WrathCombo/Combos/PvE/Content/EncounterSafety.cs
+++ b/WrathCombo/Combos/PvE/Content/EncounterSafety.cs
@@ -1,0 +1,42 @@
+#region
+
+using Dalamud.Game.ClientState.Objects.Types;
+using WrathCombo.CustomComboNS;
+using WrathCombo.Extensions;
+using static WrathCombo.CustomComboNS.Functions.CustomComboFunctions;
+using ContentInfo = ECommons.GameHelpers.Content;
+
+#endregion
+
+namespace WrathCombo.Combos.PvE.Content;
+
+/// <summary>
+///     Encounter-specific safety blocks, for enemies that must not be attacked
+///     under certain conditions.
+/// </summary>
+public static class EncounterSafety
+{
+    /// <summary>Jagd Doll's BNpcName ID.</summary>
+    private const uint JagdDoll = 3759;
+
+    private static IGameObject? Target => SimpleTarget.HardTarget;
+
+    /// <summary>
+    ///     Blocks combat actions against The Epic of Alexander (Ultimate)'s
+    ///     Jagd Dolls once they are below the 25% HP feed threshold, as any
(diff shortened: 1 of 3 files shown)


edewen:
While invincible  for auto is exactly what we would do. I'm not sure about adding a savage blade block for manual. 

Not the decider but my 2 cents is that opens up a whole can of worms that doesn't need to be opened by people who could, in fact, just stop trying to manually shoot it. Adding a whole new safety file, and in this way, as it grows would require a new line for every content specific action seems like a lot.

WrathCombo/Combos/PvE/Content/EncounterSafety.cs
```diff
@@ -0,0 +1,42 @@
...
+    ///     The action; left as the blocked action.
+    /// </param>
+    /// <returns>Whether the action should be blocked.</returns>
+    public static bool TryGetTEADollBlock(ref uint actionID)
+    {
+        if (ContentInfo.TerritoryID != 887) // The Epic of Alexander (Ultimate)
+            return false;
+
+        return Target is not null &&
+               Target.GetNameId() is JagdDoll &&
```
Taurenkey, line 39: We typically make sure we're getting the right target via Base/Data ID and not name ID matching (which according to other sources would be 11338 for the TEA variant of Jagd Doll) as to prevent issues where it's mismatched in other content, in this case normal/savage versions of the encounters these adds are lifted from.

WrathCombo/CustomCombo/Functions/Status.cs
```diff
@@ -351,6 +351,14 @@ public static bool TargetIsInvincible(IGameObject? target)
                     return true; //Unfooled means you can attack the Lyre
                 return false;
 
+            case 887: // The Epic of Alexander (Ultimate)
+                // Jagd Doll = NameId 3759
+                // Technically not invincible, but killing one wipes the raid;
+                // ignore them once below the 25% HP feed threshold
+                if (tar.GetNameId() is 3759)
```
Taurenkey, line 358: Likewise, this should be the Base/Data ID.

Taurenkey (changes requested):

PR #1192, opened 2026-06-12
[Healers] Conflicts, Bug Fixes and Tweaks, a Taurenkey inspirational PR

Taurenkey:
Similar to the WHM Cure retarget conflict, helps prevent issues with wrong retargets. This should also resolve the SGE Eukrasia spam.

Removed Eukrasia from movement on ST simple mode since it's too trigger happy.

Fix the dot action just used on check whilst I was in and refactor some redundancy out.

Add the ability to click the opener DTR to toggle the preset.

Fix MCH AoE tools.


WrathCombo/Combos/PvE/SGE/SGE.cs
```diff
@@ -505,41 +504,45 @@ protected override uint Invoke(uint actionID)
...
-            if (ActionReady(Eukrasia) && GetPartyBuffPercent(Buffs.EukrasianPrognosis) <= 50 && GetPartyBuffPercent(SCH.Buffs.Galvanize) <= 50)
-                return HasStatusEffect(Buffs.Eukrasia)
-                    ? EukrasianPrognosis
-                    : Eukrasia;
+                if (ActionReady(Pepsis) &&
+                    HasStatusEffect(Buffs.EukrasianPrognosis))
+                    return Pepsis;
+            }
+
+            if (ActionReady(Eukrasia) && GetPartyBuffPercent(Buffs.EukrasianPrognosis) <= 50 && GetPartyBuffPercent(SCH.Buffs.Galvanize) <= 50 && !HasStatusEffect(Buffs.Eukrasia) && GetPartyAvgHPPercent() > 50)
```
edewen: GetPartyAvgHPPercent() > 50

was this accidentally left greater than for testing purposes?

Taurenkey: Nope, intentional. The idea being to try and prioritise prognosis when lower health than a weaker heal that gives a shield.

edewen: 300 potency heal vs 100 potency healer + 360 potency shield. 

Gonna have to disagree with you there. Shield is the correct choice if there is no shield. Pepsis exists for that reason. But it is simple mode so it's whatever.

Taurenkey: I'll revert it, and make autorot include shields as part of the HP checks then.

edewen: advanced has shield checks calculated into hp as part of an optional thing. if you push that into autorot you invalidate peoples selections there.

Taurenkey: That's not how it works, it just means shields count towards the healing threshold so those spells have just as much impact to the way autorot works as direct healing spells.

edewen: I get what it does. a person with 65% health and a 20% shield is considered 85%. Which some people may or may not want. 

I personally don't run with shield added to hp checks on bc I want people to be topped with shields. Not hp thresholded. And adding it to autorot will now affect all 4 healers. Which is even worse. 

I get you want a solution and why. But this strips that choice from all 4 healers in advanced. 

image

edewen: Maybe that same option needs to be a single selector for simple?

zbee: > That's not how it works, it just means shields count towards the healing threshold so those spells have just as much impact to the way autorot works as direct healing spells.

So we're just not going to be healing underneath shields anymore? Just like one of the most important thing about shields, not going to do that?

Taurenkey: > So we're just not going to be healing underneath shields anymore? Just like the most important thing about regens, not going to do that?

Again, not what it means. Just now autorot looks solely at HP to determine if it needs to pivot to heals. So if you're sitting at 50% HP with a 10% shield, it thinks you have 50% HP. With the change, it will see it as 60% and if that's above the slider, no need to push more heals. If even with a shield you fall below your slider, it will still do heals. Given the shield healers will throw out shields first, it's giving them just as much value as the non-shield healers. I've been running with the change for a little while now and it's super comfy.

zbee: Yeah, no, I understood perfectly then. Shields should not count towards the threshold, all this does is help ensure that if shields crit (or a big shield ability is used) then the party will still be low when that shield ends. Shields go on top of hp, they do not wholely substitute it.
If a player is at 50% hp and get a 10% shield .... they're still 50% hp and should be healed, the shield should only really be deferring when that hp needs restored not whether it needs to be (because it does need to be).

edewen: We understand what it means. It is a blanket change affecting all healers who already have that as an option. I can only speak from a personal point, but I don't use it because I want people topped with shields not good enough with shields. 

The issue is you are stripping choices from 4 healers by adding it to auto rot instead of adding it to simple sage.

zbee: There's clearly another path forward, and that should be the one taken, because this as it is would only be suitable in very low level content (if that, probably only really for unsync stuff).

PR #1200, opened 2026-06-21
[NIN] NIN Buff Rush opener

Taurenkey:
As requested on Discord, more or less follows this https://raw.githubusercontent.com/Maplespyder/NIN-Guide/refs/heads/main/assets/images/ok.png with some liberties taken due to timing issues.


edewen:
public override List<(int[] Steps, Func<bool> Condition)> SkipSteps 
        { get; set; } = [([1,2,3], () => OriginalHook(Ninjutsu) == Suiton)];
        
This is part of the other openers, Specifically because of the timing issues with shortened opener timeout settings. It prevents the opener from going tits up if the timeout is shorter than the wait for suiton pull. Its missing from new one and needs to be dropped in there. 
        
Other than that, makes sense the catch on the step 9/10 swap. Every other opener has a fleeting immediately following its raiton somewhere without issue so i doubt propagation is an issue there.

PR #1201, opened 2026-06-21
Custom Actions

Taurenkey:
- Adds new native custom actions options to override the one button rotations.


zbee:
Without having checked the functionality at all yet: does retargeting still work if you use this?

Taurenkey:
> Without having checked the functionality at all yet: does retargeting still work if you use this?

Yes, I adjusted what was needed but I still might have missed something

edewen:
I know my big bard pr changes a LOT. so the bard seciton of this would need to be ported over to it. just on glance so far its missing that retargetting stuff on wardens paeon group cleansing. but with the big bard rewrite no point in fixing it right here

PR #1206, opened 2026-06-26
Updated Blacklist to catch some more edge cases

phecdaDia:
Current blacklist implementation appeared to not catch some cases, such as gauge based abilities (e.g. Apex Arrow)
Added explicit blacklist rules for Esuna, Kardia, should the user want them blacklisted.
Added a safeguard for when a blacklisted action is returned it'll instead return SavageBlade. 

Only caveat is blacklisted actions icons may revert to a lower level one, e.g. Refulgent Arrow shows as Straight Shot when blacklisted, some actions may not be visible at all, e.g. White Mage's Glare IV does not show instead of Presence of Mind, when it is blacklisted.

### DIFF
--- a/WrathCombo/AutoRotation/AutoRotationController.cs
+++ b/WrathCombo/AutoRotation/AutoRotationController.cs
@@ -22,6 +22,7 @@ using WrathCombo.CustomComboNS.Functions;
 using WrathCombo.Data;
 using WrathCombo.Extensions;
 using WrathCombo.Services;
+using WrathCombo.Services.ActionRequestIPC;
 using WrathCombo.Services.IPC_Subscriber;
 using WrathCombo.Window.Functions;
 using static WrathCombo.CustomComboNS.Functions.CustomComboFunctions;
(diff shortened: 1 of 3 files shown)


Taurenkey (commented):
I think the better way to do this is to implement action ready in the places it's missing, as that's the intended way to gate the actions off. Also not sure about it's implementation in auto-rotation, I want to say the original intention was for it to only affect combos and not auto-rotation but I could be wrong and I have asked for clarification.

phecdaDia:
I have changed it to ActionReady for now, however still has the issue where some actions on the hotbar are shown wrong, e.g. showing Straight Shot instead of Refulgent Arrow (Also seems to do this now when it's not even blacklisted, so not too sure why that's happening yet)

Would probably be best if it's checking the ActionReady in the individual combos then? I wanted to avoid changing stuff there since it would end up being a much larger change.

Taurenkey:
> I have changed it to ActionReady for now, however still has the issue where some actions on the hotbar are shown wrong, e.g. showing Straight Shot instead of Refulgent Arrow (Also seems to do this now when it's not even blacklisted, so not too sure why that's happening yet)
> 
> Would probably be best if it's checking the ActionReady in the individual combos then? I wanted to avoid changing stuff there since it would end up being a much larger change.

That wasn't quite what I meant. In order for this to not hang things, all implementations of the actions you're looking to blacklist in the combos themselves have to be wired up to an action ready check, there's unfortunately not lazy one-line solution to this.

phecdaDia:
Ah, my bad then 😅
I'll close this one for now then and maybe next week go through some of the actions that didn't have it wired up yet for the jobs I know. 
Thank you for your time.

PR #1207, opened 2026-06-27
BossMod Conflict Fix

Taurenkey:
Makes #1180 potentially redundant.

### DIFF
--- a/WrathCombo/Data/Conflicts/ConflictingPlugins.cs
+++ b/WrathCombo/Data/Conflicts/ConflictingPlugins.cs
@@ -463,7 +463,7 @@ public static class ConflictingPlugins
         {
             conflicts = conflicts.Append(new Conflict(
                     "BossMod", ConflictType.Settings,
-                    "AI is enabled WITH targeting [check 'Disable auto-targeting']"))
+                    "Autorotation Preset is enabled WITH targeting [remove 'Automatic Targeting' from preset or use a non-default preset]"))
                 .ToArray();
         }
 
@@ -472,6 +472,14 @@ public static class ConflictingPlugins
             conflicts = conflicts.Append(new Conflict(
                     "BossMod", ConflictType.Settings,
                     "Manual Queueing is Enabled [uncheck 'Use custom queueing']"))
+                .ToArray();
+        }
+
+        if (ConflictingPluginsChecks.BossMod.AutorotationConflicted)
+        {
+            conflicts = conflicts.Append(new Conflict(
+                    "BossMod", ConflictType.Settings,
+                    "Autorotation Preset is enabled WITH job rotations [remove rotations from preset or use a non-default preset]"))
                 .ToArray();
         }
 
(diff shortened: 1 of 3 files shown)


WrathCombo/Data/Conflicts/ConflictingPlugins.cs
```diff
@@ -463,7 +463,7 @@ private static bool TryGetSettingConflicts(out Conflict[] conflicts)
         {
             conflicts = conflicts.Append(new Conflict(
                     "BossMod", ConflictType.Settings,
-                    "AI is enabled WITH targeting [check 'Disable auto-targeting']"))
+                    "Autorotation Preset is enabled WITH targeting [remove 'Automatic Targeting' from preset or use a non-default preset]"))
```
zbee, line 466: AFAIK the previous auto-targeting setting is still functional and present in bossmod, so this isn't a 1:1 replacement.

WrathCombo/Data/Conflicts/ConflictingPluginsChecks.cs
```diff
@@ -88,7 +88,7 @@ public static void Begin()
         // ReSharper disable once RedundantAssignment
         var ts = TS.FromMinutes(1); // 1m initial delay after plugin launch
 #if DEBUG
-        ts = TS.FromSeconds(30); // 10s for debug mode
+        ts = TS.FromSeconds(1); // 10s for debug mode
```
zbee: //comment update or removal

WrathCombo/Data/Conflicts/ConflictingPlugins.cs
zbee, line 466: Now that you've added this, you haven't removed the other detection have you? Doesn't look like you did, just can't be bothered to expand every section to check.

zbee (changes requested):

zbee:
This does obsolete where I was at (locally) in #1180 yeah, looks good minus those small changes above.

WrathCombo/Data/Conflicts/ConflictingPluginsChecks.cs
zbee: You also probably just don't want it this low tbh. Don't know that that'd even work, with how the queued action checking works. If it does then whatever.

Taurenkey: It's superflous, there's another throttle further down which locks it out for 4.11s (weird number but ok)

zbee: This isn't that, but it's the initial startup delay (also not what I thought it was), so that's fine if you want it that low. Just update the comment is all I was saying.

WrathCombo/Data/Conflicts/ConflictingPlugins.cs
zbee: From dev chat convo: did remove previous check, it should be added back. But: in the mean-time it'd be entirely suitable to just add the previous setting to this comment, if it triggers this warning also.

zbee: From dev chat convo: did remove previous check, it should be added back. But: in the mean-time it'd be entirely suitable to just add the previous setting to this comment, if it triggers this warning also.

PR #1211, opened 2026-06-28
[Meta] Move custom action images

zbee:
- [X] Move Custom Action Images to where they belong
- [X] Add all relevant images to Solution Items
- [X] Add control files to Solution Items


zbee:
This part addressed the only concerns you expressed about this:
> Add all relevant images to Solution Items

It is *more* accessible than it was before.

PR #1213, opened 2026-06-29
[UI] Improve Custom Actions tab

zbee:
- [X] Removes Tables from the Custom Actions UI
      <sup>(like ... what)</sup>
- [X] Puts the Description on its own line, under the checkbox, like Presets do
- [X] Adds a helpmark with the action ID
- [X] Checks the box if the user starts dragging the icon

Before:

After:


Taurenkey:
I'm triggered 
image

zbee:
Fixed, and made descriptions grey like presets.

Kagekazu (approved):

PR #1219, opened 2026-07-03
[PLD] Added early buff opener

Lusaca87:
Adds an optional Early Buff Window opener for PLD Advanced ST, selectable next to
the Standard one (default stays Standard). It pushes Fight or Flight / Imperator
about a GCD earlier.

Hooked into the existing opener options, so the content (All/Boss Only) and
Intervene toggles work with it too. First GCD is Holy Spirit on a target, else
Fast Blade. If you're out of melee range the Atonement combo at the end is
skipped and the opener just finishes, so ranged uptime takes over.


edewen (approved):
tested and works

edewen:
Adding this in here in case in the future someone needs to refer to it. I made a graphic for it. 
image

PR #1251, opened 2026-08-09
Invincible check for Two-headed Aevis

DerLederPeter:
added an Invincibility check so it doesnt hit the immune head with the help of @Kagekazu


Kagekazu (approved):

Tartarga (approved):
As I cannot check ingame, just looking at the status ID pairing, it looks good to me

Tartarga:
This is strictly OCD purposes, and not a deal breaker, but I'd personally would like the forked tower code (1346) to be right after Clyteum (1345) to keep case numbers in numerical order

PR #1260, opened 2026-08-28
load async

Kagekazu:


Taurenkey:
I did have a go at that this last night but we still get hitches, most likely due to static classes initialising. I'll go through and see if converting most things to non-static helps.

PR #1261, opened 2026-08-28
SMN ST, and AOE were swapped in the helper file.

Genesis-Nova:
ST, and AOE's configs were swapped.
Now matching what everything else is supposed to be.


Akechi-kun (approved):

Genesis-Nova:
This will also be fixed by #1255 
Closing PR.

